PR #1, opened 2021-03-02
INIT: Project structure

aakoshh:
Created a `mill` build file with an initial jab at the project module structure.


build.sc
```diff
@@ -0,0 +1,220 @@
...
+    *
+    * Based on https://github.com/input-output-hk/iohk-monitoring-framework/tree/master/contra-tracer
+    */
+  object tracing extends SubModule with Publishing {
+    override def description: String =
+      "Abstractions for contravariant tracing."
+  }
+
+  /** Additional crypto utilities such as threshold signature. */
+  object crypto extends SubModule with Publishing {
```
KonradStaniec, line 104: I not sure what primitives we will need, but for more advanced crypto stuff we will probably need bouncy castle library

build.sc
```diff
@@ -0,0 +1,220 @@
...
+      * communicating with the interpreter over TCP.
+      */
+    object app extends SubModule {
+      override def moduleDeps: Seq[JavaModule] =
+        Seq(hotstuff.service, checkpointing.service, rocksdb, logging, metrics)
+
+      override def ivyDeps = super.ivyDeps() ++ Agg(
+        ivy"com.typesafe:config:${VersionOf.config}",
+        ivy"ch.qos.logback:logback-classic:${VersionOf.logback}",
+        ivy"io.iohk::scalanet-discovery:${VersionOf.scalanet}",
```
KonradStaniec, line 180: why the discovery is needed here ?

KonradStaniec (approved):
In general lgtm! 

Just out of curiosity, why `mill` instead of `sbt`  ?

build.sc
aakoshh: I wanted to put it here first to just create some abstractions like `Signature[OverWhatType]` and `GroupSignature[OverWhatType]` or whatever in case we need them until we can get the crypto published from Mantis. 

HotStuff recommends using k-of-n threshold signatures, so unlike the current checkpoint where the group signature is just a list, this would be single signature, but comprise of signatures from the group members. It can work both ways, but it's quicker to check just 1 signature, not 7, and requires less space.

In any case I just wanted to put it down that a module like this, shared by the others, will most likely be necessary.

build.sc
aakoshh: The current checkpointing works on a fixed topology file. I read in the analysis that this caused some problems with Nomad, where you generally don't know on what machine deployments will end up on. With discovery as long as at least 1 machine's IP is known, the rest can be configured to bootstrap from it.

I also thought it would be a good feature towards the end to allow the federation to move their machines to different IPs. 

We could enhance discovery to be able to restrict the participants by the keys they use, so a random person cannot join the discovery protocol to learn where every federation member is.

aakoshh:
> Just out of curiosity, why mill instead of sbt ?

It looks like a dubious choice, right? Especially now since the [old site](https://www.lihaoyi.com/mill/) gives 404 instead of a redirect to the [new docs on github](https://com-lihaoyi.github.io/mill/). 

I think `mill` may be a bit more discoverable, with code completion. With `sbt` I'm always hunting for something to copy paste into the DSL. And I thought Scalanet's build pipeline is a bit simpler, no Nix in sight. Maybe we can contribute a Nix exporter to `mill` at some point? ;) (I haven't checked if one exists already). Also I was too lazy to check how cross building works in `sbt`.

build.sc
KonradStaniec: on one hand it seems reasonable, on other it is just adding another vector of attack on checkpointing nodes (exposed open udp port, whole kademlia protocol running in the background etc.) 

Fixed topology seems just inherently safer and with less moving parts. 

Lets see how it will play out.

aakoshh: Yeah, it's one of the last things I thought we should do, after everything works with a fixed topology.

PR #5, opened 2021-03-09
PM-2909: Basic HotStuff

aakoshh:
Implements a generic Basic HotStuff agreement protocol as a pure state model.

Contains tests which were merged from https://github.com/input-output-hk/metronome/pull/6.

The idea is that the `ProtocolState` is a simple model that, when given an `Event` as input, returns a `(ProtocolState, Effects)` pair, which contains the next state and things that need to be done in reaction to the event. This way the model doesn't carry out any side effects. An `Effect` can be sending messages, creating blocks, or scheduling a timeout. 

The model is completely generic because I don't know yet what exact implementations to use for example for cryptographic primitives (e.g. list of signatures or a single threshold signature), and it's not necessary for the model to know. It's also generic in the type of blocks it handles, all it needs to know is how to get its hash and parent, which it receives in the form of a type class.

### Known Issues

I only started realising at the end that having an explicit `phase` and only accepting messages that are handled in that phase is forgetting that the other replicas may be slightly ahead of us. For example:
* A replica may send a `NewView` to the next leader who is not in `Prepare` yet
* Moving to `PreCommit` is triggered by receiving `Prepare`; replicas may receive it earlier and cast their votes, which might reach the leader while it's still in `Prepare`.
* Similarly, moving to next phases are triggered by the receipt of a Q.C., and again the leader might get the vote too early.
* In a federation of 10, the first 7 members could already have done the Prepare phase of view N, and the leader sent the new block to the last 3 who are for some reason still in the Decide phase of the previous view N-1, and not expect them

There are multiple possible approaches to work around these issues:
1. We can inspect unexpected messages to see if they are "in the future" compared to what we're expecting right now, and re-deliver them after a transition has been made. To help with this, messages like these are wrapped in a `TooEarly` error.
2. We can get rid of the explicit `phase` and accumulate all votes in different phase-buckets during the view, but this doesn't handle the case of messages arriving early across views.
3. We can make sure the leader processes `SendMessage(self, _)` effects first, so any state transitions are carried out before messages are sent out, to minimise mismatches (although processing local messages are anyway likely to be quicker than a network roundtrip).

It sounds reasonable to leave the handling of unexpected messages to the host component rather than try to manage stashing in the `ProtocolState`: if the message is just from the next round, we can re-deliver it as soon as the view number increases after an event; if, however, the messages are further from the future, we could trigger synchronisation with the other peers instead. The re-queueing could also reason about which messages are worth keeping (`NewView`, `Prepare`, `Quorum`). The host system has to do block synchronisation and validation anyway, so it has to inspect the messages.

### Host System

The intention is that the `ProtocolState` will be embedded in an effectful host system, which sets up a pipeline:
* network events come in and go through stateless validation, then enter the synchronisation queue 
* events in the sync queue get their dependencies resolved in go in the PoW validation queue
* events in the PoW validation queue get validated against the interpreter, then added to the consensus queue
* events in the consensus queue get run against the protocol state:
  * if the result is `TooEarly` they get in a stashed 
  * if a transition was made then previously stashed events get unstashed (up to where they were expected)
  * messages to self are selected and executed in a loop, advancing the state further and yielding further effects
  * the accumulated effects are forked to be executed in parallel
  * the next event from the consensus queue will use the updated state

### Questions

1. The model doesn't accept a `Vote` or a `Quorum` that is about a different block hash than the the one which was `Prepare`d in this view. This isn't part of the HotStuff paper pseudo-code; there every Vote and Quorum Certificate contains a block directly and it's not checked that it's the same block as the previous one. So for example if the leader proposes block `A`, but all votes that come back are about block `B`, it will produce a Q.C. for `B`. Or if a replica receives a `Prepare` with block `A` from the leader, and then a Q.C. for block `B` it will vote on `B`. I didn't think these make too much sense in a scenario where we have n-f honest nodes, so it is not allowed at the moment. But the paper says _it introduces a second phase that allows replicas to “change their mind” after voting in the phase_, so I'm not sure that I'm not missing something.
2. The pseudo-code always extends the `highQC` by just one block, where the `highQC` is the highest `prepareQC` from the previous round, as reported by replicas; the safety checks, however, don't really specify what "extends from" means, whether it has to be a direct descendant or there can be a chain. I defined it as a direct descendant because that is how the block creation works, and because this was easier to check since there's no need to access any chain history for it. Happily this means that if a Byzantine leader sends us a block to prepare, it should never force us to have to download a large chain of missing blocks in futility, because the Q.C. that accompanies the prepare message should prove that the parent of the block was agreed by the federation, so it should be okay to download. Looking at its view number, we should be able to tell whether it's in the past, and irrelevant (older than our committed Q.C.) or far in the future (which means we've fallen behind and could potentially skip ahead).


metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/Federation.scala
```diff
@@ -0,0 +1,20 @@
...
+  /** Size of the federation, `n`. */
+  val size: Int = publicKeys.size
+
+  /** Maximum number of Byzantine nodes, `f`, so that `n >= 3*f+1`. */
+  val maxFaulty: Int = (size - 1) / 3
+
+  def contains(publicKey: PKey): Boolean =
+    publicKeySet.contains(publicKey)
+
+  def leaderOf(viewNumber: ViewNumber): PKey =
```
ranvirranaiitb, line 18: Leader assignment based on view-number has not been discussed in the Hotstuff paper and in general, it does not affect the safety and liveness. However, it does affect worst-case latency.

Consider a static adversary under a round-robin leader change scheme. All the f nodes can set their public keys so that they are consecutive. In such a scenario those f consecutive leaders can create timeouts leading to an O(f) confirmation latency. (Recall that in a normal case, the latency is O(1)). 

A minor improvement to this is to assign leaders based on "publicKeys((H256(viewNumber).toInt % size).toInt)". This leader order randomization via a hash function will ensure that even if adversarial public keys are consecutive in PublicKey set, they are not necessarily consecutive in leader order. 

Not that the above policy will not ensure that adversarial leaders are never consecutive, but the probability of such occurrence will be lower under a static adversary.

PS: I used "H256(viewNumber).toInt" as shorthand for some concatination of the hash

ranvirranaiitb:
Hi Akosh,

Great implementation of the HotStuff Protocol State model. I have left a few comments. 
Regarding the two questions, here are my thoughts:

1.) I don't think we need to change the current model. If the leader proposes Block `A`, all honest nodes will vote on `A` as described in Hotstuff paper Algorithm 2 line 10. ("send voteMsg(prepare, m.node , ⊥) to leader(curView )"). Hence we should not allow votes on `B`. However, we don't need to explicitly write validation where a leader checks if the vote cast is on `A`; this is because a QC can only be created for block `A` (assuming n-f honest). 
The statement " “change their mind” after voting in the phase" probably refers to the differentiation of a three-phase protocol with two-phase protocols. In two-phase protocols, locked QC is generated in the second phase hence the replicas "can't change their mind after voting", whereas in Hotstuff lockedQC is generated in the third phase. This delayed lock allows three-phase protocols to change view quickly as compared to two-phase protocols. Two phase-protocols require the leader to wait for \Delta to receive all honest votes to determine HighQC. 

2.) Just extending the high QC leaf in the current tree by one node should be good enough. I don't see any advantage in extending the tree using a chain of nodes, especially if the chain is created by the same leader and broadcast at the same time.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
```diff
@@ -0,0 +1,492 @@
...
+              val next = moveTo(Phase.Decide).copy(
+                lockedQC = m.quorumCertificate
+              )
+              next -> effects
+            }
+        }
+
+      // Leader:  Collect Commit votes, broadcast Commit Q.C.
+      // Replica: Wait for Commit Q.C., execute block, send NewView, move to Prepare.
+      case Phase.Decide =>
```
span14, line 233: I think since we are not doing Chained Hotstuff, during the decision phase, we should clear out variables like lockedQC, etc. used in the previous view.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
```diff
@@ -0,0 +1,492 @@
...
+    next -> effects
+  }
+}
+
+object ProtocolState {
+
+  /** The result of state transitions are the next state and some effects
+    * that can be carried out in parallel.
+    */
+  type Transition[A <: Agreement] = (ProtocolState[A], Seq[Effect[A]])
```
span14, line 462: I don't have much experience coding with Scala and I am curious about the architecture for Hotstuff in general. Based on the code here, in my opinion, there are two threads running in parallel with one running the protocolState update and the other handling effects in sequence. If that's the case, how to process some effects like ScheduleNextView, which correlates with protocolState update as well, and shall we do it synchronously or asynchronously.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
```diff
@@ -0,0 +1,492 @@
...
+  import Effect._
+  import Event._
+  import ProtocolState._
+  import ProtocolError._
+
+  val leader   = federation.leaderOf(viewNumber)
+  val isLeader = leader == publicKey
+
+  /** The leader has to collect `n-f` signatures into a Q.C. */
+  def quorumSize = federation.size - federation.maxFaulty
```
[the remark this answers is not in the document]
aakoshh: Thank you for that explanation! I did it this way to follow the paper, but I also thought that it was a bit unintuitive that if we assume n=10 and f=1, then we have to get a quorum of 9, while f=3 requires (or allows) a smaller quorum size. Seeing the whole inequality makes more sense.

I thought this could be a separate PR, but on second thought I'll put it into this one, since `maxFaulty` can be set to 1 and then it doesn't make any sense the way it is.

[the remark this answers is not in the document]
aakoshh: Ah, forgot that `maxFaulty` is not something you can set, it always uses the most pessimistic value, giving the largest quorum.

In that light, maybe a separate PR _is_ better. Do you see `f` being configurable?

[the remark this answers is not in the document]
aakoshh: Created https://jira.iohk.io/browse/PM-3058

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
```diff
@@ -0,0 +1,492 @@
...
+    * because the leader is a replica as well. The effect
+    * system should take care that these messages don't
+    * try to go over the network.
+    *
+    * NOTE: Some messages trigger transitions; it's best
+    * if the message sent to the leader by itself is handled
+    * before the other messages are sent out to avoid any
+    * votes coming in return coming in phases that don't
+    * yet expect them.
+    */
```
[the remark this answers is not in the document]
aakoshh: Exactly! I made this comment before I added `TooEarly` to help with this buffering, but then I thought it probably makes sense anyway to carry out local effects first, then we're double protected. It can't handle the case of `NewView` arriving from a replica that timed out, so `TooEarly` seems necessary, but if it's easy to do there wouldn't be any harm in it, it would just act on things we already know, and most likely it wouldn't delay sending out messages by much. 

We'll see, if it does, we can always just rely on buffering, but at least the comment can remind us of the option.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/Federation.scala
aakoshh: Thank you for your input, I'll add it to https://jira.iohk.io/browse/PM-2938

I read that Libra uses a Verifiable Random Function, and yours makes sense as well, but there are some potential issues such as: what if we repeatedly pick a leader that is offline for example? Maybe we should do a short simulation to see how often `H256(viewNumber).toInt % size` tends to bounce around in a small clique? 

Or perhaps have some function that uses some permutation of the leaders in round-robin, with deterministic rotation after each cycle?

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
aakoshh: So you're saying that if we reached the decide phase in view N, then the `lockedQC` of view `N` should be forgotten? And by "like" you mean `prepareQC` as well?

There's no other "previous view", there is just 1 `lockedQC`. If we clear it out, then in view `N+1` we can't use it in the `isSafe` method to check that the block proposed by the next leader extends it. We also cannot clear out `prepareQC` because if `N+1` times out, then it's the only `prepareQC` we can send to the leader of `N+2`.

Or did I misunderstand your suggestion?

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
aakoshh: Thank you for the question! I tried to describe on a high level under the "Host System" section of the PR description.

The idea is that there would be 1 fiber going through the `Event`s in a `ConcurrentQueue`, where every event can create a new `ProtocolState` and some `Effect`s. The effects can be handled concurrently, they don't have to be in sequence; for example we can send messages to all federation members in parallel. If an `Effect` results in a new `Event` (such as a `CreateBlock` eventually produces a `BlockCreated`), that goes in the event queue, to be processed in sequence. Similarly, the host system can carry out `ScheduleNextView` in the background and put a `NextView` event on the queue when its' done:

```scala
def startProcessingEffect(effect: Effect): F[Fiber[F, Unit]] = effect match {
  case ScheduleNextView(viewNumber, timeout) = 
    Concurrent[F].start {
      Timer[F].sleep(timeout) >> eventQueue.offer(Event(NextView(viewNumber))
    }
  ...
}
```

Here's a [blog](https://dev.to/kspeakman/testable-back-end-programming-in-f-41n5) that illustrates this design beautifully.

aakoshh:
Thank you for answering my questions @ranvirranaiitb 

Looks like there's no need to change anything then. I'd leave in the checks for the block hash - it doesn't do any harm, but it has to be tested by at least the leader at some point, so it saves some extra work and testing.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
span14: Yeah, I misunderstood the protocol and there is no need to clear out both QCs in this case.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
span14: Nice introduction. I will take a look at the blog.

aakoshh: Here are links to the classes mentioned: 
* [ConcurrentQueue](https://monix.io/api/3.1/monix/catnap/ConcurrentQueue.html)
* [Ref](https://typelevel.org/cats-effect/concurrency/ref.html)
* [Timer](https://typelevel.org/cats-effect/datatypes/timer.html)
* [Fiber](https://typelevel.org/cats-effect/datatypes/fiber.html)
* [Concurrent](https://typelevel.org/cats-effect/typeclasses/concurrent.html)
* [cats-effect](https://typelevel.org/cats-effect/api/cats/effect/index.html) API docs in general

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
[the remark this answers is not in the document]
ranvirranaiitb: I think we should fix `f = n/3-1`. That way both n-f and (n+f)/2 would be the same.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/Federation.scala
ranvirranaiitb: `H256(viewNumber).toInt % size` will not bounce around a small clique since H256 can model a uniform distribution. (% size) will remove some uniformity but not by observable standards. 

"repeatedly pick a leader that is offline for example": This can happen even in round-robin or VRF. 

The problem with `H256(viewNumber).toInt % size` is that future allotment is predictable. The adversary can see which public key is going to be a leader at a target viewNumber and can choose to corrupt it or DoS it. 

VRF is definitely superior since the future allotment is not predictable. However, it will take more time to implement and can be implemented at a later stage.

ranvirranaiitb (approved):

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/Federation.scala
aakoshh: Yes, picking the same guy can definitely happen with the VRF, but I thought with round-robin at least we only hit someone offline in every Nth round. 

I agree the hash should be uniform, happy to go with that to randomise the order.

metronome/hotstuff/consensus/src/metronome/hotstuff/consensus/basic/ProtocolState.scala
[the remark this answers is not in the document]
aakoshh: Hm, that's kind of what's [happening now](https://github.com/input-output-hk/metronome/pull/5/files#diff-d56a67191a6bf58e5b4c7b05b74eed1db2fc992b462b4a77dda908974f645e64R12-R13), resulting both upper and lower thresholds be the same, albeit with `f = (n-1)/3` (so with 10 nodes it gives (10-1)/3=3, not 10/3-1=2). 

I thought the benefit of your formula would be that we could set an `f` which is lower than the theoretical maximum, and thus be content with a lower quorum size, e.g. set f=1 and get (10+1)/2+1=6 instead of 7.

PR #12, opened 2021-03-26
PM-3058: Allow specifying the number of Byzantine nodes.

aakoshh:
Add a constructor to `Federation` that allows setting the `f` to a value lower than the theoretical maximum, resulting in lower quorum size.


ranvirranaiitb (approved):
Hi Akosh,

Thanks for the update. 
I saw that you have separated maxFaulty and f. This will add to the flexibility, however, it should be noted that ideally, we should keep maxFaulty = function(n). Higher n will imply more nodes can be faulty. 
This is not a problem since we only need to set these variables once, we can manually ensure that maxFaulty is reasonably large (as compared to n).

aakoshh:
> noted that ideally, we should keep maxFaulty = function(n).

@ranvirranaiitb I agree, and this is what the [default constructor](https://github.com/input-output-hk/metronome/pull/12/files#diff-d56a67191a6bf58e5b4c7b05b74eed1db2fc992b462b4a77dda908974f645e64R48-R52) achieves. Specifying `f` manually is optional.

I hope I haven't misunderstood things. My understanding is that the `q = (n + f) / 2 + 1` formula allows a `q < n - f` _iff_ `f < (n - 1) / 3`, so the idea is that if we think the federation is trustworthy then we can configure `f` manually and enjoy a bit more speed, but it's not mandatory to do so, in which case it should use the default, resulting in `q = n - f`. 

Currently this manual setting would have to be applied uniformly on all nodes as well, because quorum certificates are only accepted if they contain _exactly_ the expected number of signatures (at least in the tests). I guess that makes sense because if Alice is happy with 6 signatures while Bob wants 7, Bob won't accept a Q.C. made by Alice anyway, so it's not a problem that Alice would reject Bob's too. Also, if we used threshold signatures, Alice wouldn't even be able to produce a valid group signature based on her own preferences with fewer than `k` signatures.

PR #13, opened 2021-03-29
PM-2907: Emit trace events from networking.

aakoshh:
Adds structured logging to capture some network events via tracing. The logs are only demonstrated in the tests (and only visible if the log level in `networking/test/resources/logback.xml` is adjusted from `ERROR` to `DEBUG`).

Adopted a [hybrid log format](https://medium.com/unomaly/logging-wisdom-how-to-log-5a19145e35ec) consisting of a human readable `message` and key-value pairs in the form of a JSON object in `event`.

The resulting logs look like this:
```
21:25:44.871 DEBUG i.i.m.networking.NetworkEvent ConnectionFailed {"key":"1bc14e1ac7f929c83b05d796a591c272e36ab7a6e71df41c330c1e21d377af6d300cb73a9395da68708fc2d05c0cd1800b4d07dcf88940f05b161277da4926cd","address":"localhost/127.0.0.1:45765","err":"Error establishing channel to PeerInfo(BitVector(512 bits, 0x1bc14e1ac7f929c83b05d796a591c272e36ab7a6e71df41c330c1e21d377af6d300cb73a9395da68708fc2d05c0cd1800b4d07dcf88940f05b161277da4926cd),localhost/127.0.0.1:45765)."}
21:25:44.905 DEBUG i.i.m.networking.NetworkEvent ConnectionRegistered {"key":"98e7fdb3b38f7a15ec8f9feb407460996db4b075869930571503ba47af0a96341c8d07c97031b839874451a748a0d28b35f47f7a53f22a086eba5045b0a74d08","address":"localhost/127.0.0.1:46251"}
```

There's also a class to help collect the logs in memory, so we can assert for example that no warnings have been emitted, or set up multiple nodes in integration tests and add their hostname to the logs on the fly, which can help debugging issues. 

`NetworkTracers` is a case class so we can individually compose tracers for separate events using `copy`, for example to disable one by replacing it with `Tracer.noOpTracer` or later to enhance by adding metrics, something like this:

```scala
implicit val networkTracers: Tracer[F, NetworkEvent[K]] = {
  val log = NetworkTracers(LogTracer.hybrid[F, NetworkEvent[K]])
  val registered = Metrics.gauge("connections")
  log.copy(
    registered = log.registered |+| Tracer.const(Sync[F].delay(registered.increment())),
    deregistered = log.deregistered |+| Tracer.const(Sync[F].delay(registered.decrement()))
  )
}
```

Silenced logs from Scalanet in tests.

### DIFF
--- a/metronome/logging/src/io/iohk/metronome/logging/HybridLog.scala
+++ b/metronome/logging/src/io/iohk/metronome/logging/HybridLog.scala
@@ -1,0 +1,33 @@
+package io.iohk.metronome.logging
+
+import io.circe.JsonObject
+import java.time.Instant
+import scala.reflect.ClassTag
+
+trait HybridLog[T] {
+  def apply(value: T): HybridLogObject
+}
+
+object HybridLog {
+  def apply[T](implicit ev: HybridLog[T]): HybridLog[T] = ev
+
+  def instance[T: ClassTag](
+      level: T => HybridLogObject.Level,
+      message: T => String,
+      event: T => JsonObject,
+      source: Option[String] = None
+  ): HybridLog[T] =
+    new HybridLog[T] {
+      val className = implicitly[ClassTag[T]].runtimeClass.getName
+
+      override def apply(value: T): HybridLogObject = {
+        HybridLogObject(
+          level = level(value),
+          timestamp = Instant.now(),
+          source = source getOrElse className,
+          message = message(value),
+          event = event(value)
+        )
+      }
+    }
+}
--- a/metronome/logging/src/io/iohk/metronome/logging/HybridLog.scala
+++ b/metronome/logging/src/io/iohk/metronome/logging/HybridLog.scala
@@ -4,6 +4,7 @@ import io.circe.JsonObject
 import java.time.Instant
 import scala.reflect.ClassTag
 
+/** Type class to transform instances of `T` to `HybridLogObject`. */
 trait HybridLog[T] {
   def apply(value: T): HybridLogObject
 }
@@ -11,20 +12,22 @@ trait HybridLog[T] {
 object HybridLog {
   def apply[T](implicit ev: HybridLog[T]): HybridLog[T] = ev
 
+  /** Create an instance of `HybridLog` for a type `T` by passing
+    * functions to transform instances of `T` to message and JSON.
+    */
   def instance[T: ClassTag](
       level: T => HybridLogObject.Level,
       message: T => String,
-      event: T => JsonObject,
-      source: Option[String] = None
+      event: T => JsonObject
   ): HybridLog[T] =
     new HybridLog[T] {
-      val className = implicitly[ClassTag[T]].runtimeClass.getName
+      val source = implicitly[ClassTag[T]].runtimeClass.getName
 
       override def apply(value: T): HybridLogObject = {
         HybridLogObject(
           level = level(value),
           timestamp = Instant.now(),
-          source = source getOrElse className,
+          source = source,
           message = message(value),
           event = event(value)
         )
--- a/metronome/logging/src/io/iohk/metronome/logging/LogTracer.scala
+++ b/metronome/logging/src/io/iohk/metronome/logging/LogTracer.scala
@@ -1,0 +1,1 @@
+// package io.iohk.metronome.logging
--- a/metronome/logging/src/io/iohk/metronome/logging/LogTracer.scala
+++ b/metronome/logging/src/io/iohk/metronome/logging/LogTracer.scala
@@ -1,1 +1,38 @@
-// package io.iohk.metronome.logging
+package io.iohk.metronome.logging
+
+import cats.syntax.contravariant._
+import cats.effect.Sync
+import io.circe.syntax._
+import io.iohk.metronome.tracer.Tracer
+import org.slf4j.LoggerFactory
+
+/** Forward traces to SLF4J logs. */
+object LogTracer {
+
+  /** Create a logger for `HybridLogObject` that delegates to SLF4J. */
+  def hybrid[F[_]: Sync]: Tracer[F, HybridLogObject] =
+    new Tracer[F, HybridLogObject] {
+      override def apply(log: => HybridLogObject): F[Unit] = Sync[F].delay {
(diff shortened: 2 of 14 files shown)


metronome/networking/test/src/io/iohk/metronome/networking/RemoteConnectionManagerWithScalanetProviderSpec.scala
```diff
@@ -129,9 +132,40 @@ object RemoteConnectionManagerWithScalanetProviderSpec {
...
+      : NetworkTracers[F, K, M] = {
+    import io.circe.syntax._
+    import NetworkEvent._
+
+    def peerJson(key: K, address: InetSocketAddress) =
+      JsonObject("key" -> key.asJson, "address" -> address.toString.asJson)
+
+    implicit val hybridLog: HybridLog[NetworkEvent[K]] =
+      HybridLog.instance[NetworkEvent[K]](
+        level = _ => HybridLogObject.Level.Debug,
```
KonradStaniec, line 149: Just to make sure if I understand correctly, If I would like to have differnt log levels for different event I would need to patter match them here for attach correct log levels to them ?

metronome/logging/src/io/iohk/metronome/logging/LogTracer.scala
```diff
@@ -0,0 +1,38 @@
...
+  def hybrid[F[_]: Sync]: Tracer[F, HybridLogObject] =
+    new Tracer[F, HybridLogObject] {
+      override def apply(log: => HybridLogObject): F[Unit] = Sync[F].delay {
+        val logger = LoggerFactory.getLogger(log.source)
+
+        def message = s"${log.message} ${log.event.asJson.noSpaces}"
+
+        log.level match {
+          case HybridLogObject.Level.Error =>
+            if (logger.isErrorEnabled) logger.error(message)
```
KonradStaniec, line 22: If we would https://github.com/lightbend/scala-logging here, we could probably avoid those `logger.isLevelEnabled` checks

metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
```diff
@@ -28,14 +28,15 @@ class ConnectionHandler[F[_]: Concurrent, K, M](
...
-) {
+)(implicit tracers: NetworkTracers[F, K, M]) {
 
   private val numberOfRunningConnections = AtomicInt(0)
 
   private def closeAndDeregisterConnection(
       handledConnection: HandledConnection[F, K, M]
   ): F[Unit] = {
     for {
+      _ <- tracers.deregistered(handledConnection)
```
KonradStaniec, line 39: minor thing, I would add `deregistered` event as  last thing or maybe as `closeAndDeregisterConnection` finalizer i.e
`closeAndDeregisterConnection(connection).guarantee(tracers.deregistered(handledConnection))`

metronome/networking/src/io/iohk/metronome/networking/NetworkEvent.scala
```diff
@@ -0,0 +1,28 @@
+package io.iohk.metronome.networking
+
+import java.net.InetSocketAddress
+
+/** Events we want to trace. */
+sealed trait NetworkEvent[K]
```
KonradStaniec, line 6: Maybe another useful ones would be: `MessageReceived` and `ConnectionError` ?

metronome/networking/test/src/io/iohk/metronome/networking/RemoteConnectionManagerWithScalanetProviderSpec.scala
aakoshh: Yeah, you can issue a warning/error for certain types if you want to.

metronome/logging/src/io/iohk/metronome/logging/LogTracer.scala
aakoshh: Yeah but it does it in the background via macros. I thought using this directly we can avoid the extra machinery that puts in line numbers and source classes or whatever, because they are irrelevant/unknown.

metronome/networking/src/io/iohk/metronome/networking/NetworkEvent.scala
aakoshh: Is `ConnectionError` different from `ConnectionFailed`?

Sure, I'll add `MessageReceived` and `MessageSent` as well.

KonradStaniec: `ConnectionFailed` is when we call to remote peer and fail,  and `ConnectionError` would be when we are listening for incoming event from peer and receive lets say `DecodingError`

metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
aakoshh: Done.

metronome/networking/src/io/iohk/metronome/networking/NetworkEvent.scala
aakoshh: Done, added `ConnectionReceiveError`, `ConnectionSendError`, `ConnectionUnknown`, `MessageSent` and `MessageReceived`.

KonradStaniec (approved):
this will be terribly helpful to know what is happening on network level ! 👍

metronome/logging/src/io/iohk/metronome/logging/HybridLog.scala
```diff
@@ -0,0 +1,36 @@
...
+  /** Create an instance of `HybridLog` for a type `T` by passing
+    * functions to transform instances of `T` to message and JSON.
+    */
+  def instance[T: ClassTag](
+      level: T => HybridLogObject.Level,
+      message: T => String,
+      event: T => JsonObject
+  ): HybridLog[T] =
+    new HybridLog[T] {
+      val source = implicitly[ClassTag[T]].runtimeClass.getName
```
lemastero, line 24: 👍

PR #15, opened 2021-03-30
PM-2929: Transaction data structure.

aakoshh:
Adds the most minimalistic `Transaction` type I can think of. 

I think these can become part of the ledger as-is. 

So far we haven't discovered a need for the service to know the `height` of the proposer blocks and the checkpoint candidate. I thought this could be relevant during the update of the ledger: say we have to checkpoint every 10th height in the chain, there's no checkpoint yet for height 100, and we have these events:
1. Alice adds a proposer block transaction for level 101
1. Bob adds a checkpoint candidate for level 100 - it probably cannot enforce the inclusion of Alice's proposer at level 101
1. The ledger is updated, the accumulated proposer blocks are discarded, the new checkpoint is the one sent by Bob - should we keep Alice's proposer block in the ledger?

During our discussions I think @ranvirranaiitb said that if Alice was honest she'd have added a checkpoint candidate for level 100 if she has something to say about level 101. In that case perhaps she could have got the checkpoint in, clear the ledger, then process the next transaction which adds the new proposer at level 101, all without the checkpoint service having to know the heights. 

Another alternative is to return in a response from the interpreter which proposer block entries can be cleared from the ledger, after the execution of a block. 

Other fields that came to mind are timestamp, or signatures, but it doesn't seem like there's any use for those. Not even the block needs to be signed, since it can only arrive in a Prepare message from the leader, and after that everyone signs it via Votes. 

I thought about making this generic, so in the service side it can be something like `CheckpointCandidate[BitVector]`, while on the interpreter side it can be a `CheckpointCandiate[Vector[VoterBlock.Hash]]`. Not sure if that would make it easier or more difficult to understand: the idea is that the `checkpointing.interpreter` module is a library that is independent of the PoW domain, so for this interpreter it's opaque, and that's emphasized by the fact that it's just a `BitVector`. The raw data has to be packed/unpacked explicitly by the ETC interpreter.

### DIFF
--- a/metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/models/Transaction.scala
+++ b/metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/models/Transaction.scala
@@ -1,0 +1,54 @@
+package io.iohk.metronome.checkpointing.interpreter.models
+
+import scodec.bits.BitVector
+
+/** Transactions are what comprise the block body used by the Checkpointing Service.
+  *
+  * The HotStuff BFT Agreement doesn't need to know about them, their execution and
+  * validation is delegated to the Checkpointing Service, which, in turn, delegates
+  * to the interpreter. The only component that truly has to understand the contents
+  * is the PoW specific interpreter.
+  *
+  * What the Checkpointing Service has to know is the different kinds of transactions
+  * we support, which is to register proposer blocks in the ledger, required by Advocate,
+  * and to register checkpoint candidates.
+  */
+sealed trait Transaction
+
+object Transaction {
+
+  /** In PoW chains that support Advocate checkpointing, the Checkpoint Certificate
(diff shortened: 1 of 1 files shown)


ranvirranaiitb (approved):
I think it's a good idea to make those transactions minimalistic. 

The service does not need to store `height` since all the validation and block creation tasks are performed at the Checkpointing interpreter. The CP interpreter can map those transactions (decoded to proposerHashes, candidates) to their respective block height (by asking ETC).

aakoshh:
> Which module do you think should be responsible for creating HotStuff blocks? Checkpointing service, Checkpointing Interpreter or ETC? That module should know the height.

The HotStuff blocks would be created by the Checkpointing Service in my view, it's the one that knows what content goes into the block. The interpreter would only be responsible for producing and validating the transactions. So I thought not even the service needs to know the height, but if that turns out to be false, we can add it later.

PR #17, opened 2021-03-30
PM-2931: Ledger data structure

aakoshh:
Defines a simple `Ledger` type with the last checkpoint (can be empty in the beginning) and the concatenation of proposer blocks. The `update` function clears the proposer blocks if there's a new checkpoint to replace the old one.

Added RLP codecs for calculating the hash, which we can make part of the blocks.


ranvirranaiitb (approved):
I think the ledger state is containing only unreferenced proposers and checkpoint candidate should be sufficient. In the future, we might need to store the votes for the prepareQC and preCommitQC for past views somewhere (not necessarily in the ledger) for use in Forensics.

PR #18, opened 2021-03-31
PM-2930: Block data structure

aakoshh:
Creates a `Block` data structure with a `Block.Header` and `Block.Body` part. I thought it would have actually eliminated some possible hash mismatches if the header and body wasn't separated, but in the concept document the working theory was that the Checkpoint Certificate will have a list of block headers in it, so I stick to it. I guess it's conventional as well to separate the two.

Introduced separate `Hash` types for each of `Header`, `Body` and `Ledger`. 

__Questions__

Do we want to add more fields to the `Block.Header`?
* `height`: When someone proposes a block to us, and we miss its parent, we could look at the last block we have a cetificate for and ask for a whole range of blocks, rather than one by one, following parent-child relationships. It could also help with pruning old blocks if we had stored them by height.
* `viewNumber`: Every `Propose` message has a block in it that matches its view number. If we're missing blocks, we could follow the parents and make sure that every block along the way was created by a federation member that lead that view; however, you can only extend blocks that have a _Prepare Q.C._ so we shouldn't have a chain of blocks that aren't supported by some quorum along the way, which has the `viewNumber` in it.
* `creatorPublicKey`: Shouldn't be necessary if we can show a _Prepare Q.C._ that signed the `blockHash`.
* `creatorSignature`: It could support the claim that the block matches what the creator proposed, but it shouldn't be necessary as long as a _Prepare Q.C._ is available. And since we receive the block right from the leader in `Prepare`, over an authenticated channel, there's no need to sign the content.

We could also remove `preStateHash` to save some space, it's just a copy of the `postStateHash` from the parent of the block.

__Note__

I noticed in https://github.com/input-output-hk/metronome/pull/20 that these models have to be visible to the interpreter and the PoW chain after all, because they are part of the Checkpoint Certificate, to provide proof of the BFT agreement. The code has been moved around in that PR.

### DIFF
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
@@ -34,6 +34,11 @@ object Block {
 
   case class Header(
       parentHash: Hash,
+      // Hash of the Ledger before executing the block.
+      preStateHash: Hash,
+      // Hash of the Ledger after executing the block.
+      postStateHash: Hash,
+      // Hash of the transactions in the body.
       bodyHash: Hash
   ) extends RLPHash[Header] {
     protected override val encoder = RLPCodecs.rlpBlockHeader
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
@@ -6,6 +6,13 @@ import io.iohk.metronome.checkpointing.interpreter.models.Transaction
   * with the transactions in the body being the "commands".
   *
   * The block contents are specific to the checkpointing application.
+  *
+  * The header and body are separated because headers have to part
+  * of the Checkpoint Certificate; there's no need to repeat all
+  * the transactions there, the Merkle root will make it possible
+  * to prove that a given CheckpointCandidate transaction was
+  * indeed part of the block. The headers are needed for parent-child
+  * validation in the certificate as well.
   */
 sealed abstract case class Block(
     header: Block.Header,
@@ -39,6 +46,10 @@ object Block {
       postStateHash: Ledger.Hash,
       // Hash of the transactions in the body.
       bodyHash: Body.Hash
+      // TODO (PM-3102): Add merkle root for contents.
+      // Instead of the hash of the body, should we use the
+      // the Merkle root of the transactions?
+      // Or should that be an additional field?
   ) extends RLPHash[Header, Header.Hash]
 
   object Header extends RLPHashCompanion[Header]()(RLPCodecs.rlpBlockHeader)
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
@@ -40,8 +40,6 @@ object Block {
 
   case class Header(
       parentHash: Header.Hash,
-      // Hash of the Ledger before executing the block.
-      preStateHash: Ledger.Hash,
       // Hash of the Ledger after executing the block.
       postStateHash: Ledger.Hash,
       // Hash of the transactions in the body.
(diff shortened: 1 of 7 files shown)


metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/models/Block.scala
```diff
@@ -0,0 +1,62 @@
...
+    */
+  def make(
+      header: Header,
+      body: Body
+  ) = makeUnsafe(header = header.copy(bodyHash = body.hash), body = body)
+
+  case class Header(
+      parentHash: Header.Hash,
+      // Hash of the Ledger before executing the block.
+      preStateHash: Ledger.Hash,
```
ranvirranaiitb, line 44: Why do we need preStateHash, isn't preStateHash = parent.postStateHash?

aakoshh: It is, in fact I added a [better constructor](https://github.com/input-output-hk/metronome/pull/20/files#diff-b97700721336d64d734806933d3aa6a8e188fd821718570e27f71df253791a35R35-R50) to `Block` that takes the parent and just copies this. 

We don't need it, probably, I added because I once worked on a project that had it like this. I'll remove it, we can always add it if it's useful.

ranvirranaiitb:
I think adding more fields to the block header is a good idea. Although I don't see any use for `height` and `viewNumber`, it might be helpful in the future, (say to easily handle some scenarios with a dynamic validator set). 

Very good question regarding `creator.pubKey` and `creator.signature`, I was wondering if we allow the possibility of Man-in-the-middle attacks in authenticated channels. We probably don't but if we do, it's important to include `creator.signature` to prevent an adversary from changing block in everyone's prepare messages and causing a liveness failure. Perhaps a better location to place this signature and pubkey is in the prepare message. (This will also allow for gossip/routing if it's needed in the future).

aakoshh:
> I was wondering if we allow the possibility of Man-in-the-middle attacks in authenticated channels. 

According to @KonradStaniec the way the `DynamicTlsPeerGroup` establishes connections is to carry out a verification of the key the other side uses, following the libp2p protocol specification. This key is independent from what's securing the channel itself as I understand though, so maybe a MITM attack would be possible if we have `Alice <-> Eve <-> Bob` and Eve can forward all messages that, say, Alice signs with her private key, in response to a challenge from Bob. But that's so obvious that I'd assume that the protocol should protect against it? If it were true, though, then indeed Eve could later modify a block that Alice is sending.

@KonradStaniec can you help with this question?

>  Perhaps a better location to place this signature and pubkey is in the prepare message. (This will also allow for gossip/routing if it's needed in the future).

Yes that would be an easy extension. I was wondering if block signatures could protect against manipulation during syncing past history; maybe we can even use the historic `Prepare` messages for sync.

aakoshh:
Here's the [spec](https://github.com/libp2p/specs/blob/master/tls/tls.md) and the [DynamicTLSExtension](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/src/io/iohk/scalanet/peergroup/dynamictls/DynamicTLSExtension.scala) that refers to it.

aakoshh:
> The peer signs the concatenation of the string libp2p-tls-handshake: and the public key that it used to generate the certificate carrying the libp2p Public Key Extension, using its private host key. This signature provides cryptographic proof that the peer was in possession of the private host key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.

My understanding is that the key extension contains a signature over the X.509 public key using the host key (i.e. the Secp256k1 private key used for federation membership), so it proves that the connection is secured with a certificate that the federation member created itself, and not by a man-in-the-middle.

aakoshh:
NB: There are only a few elliptic curves supported by this scheme, so if we want to switch to k-of-n signatures with partial signatures on votes, we may have to think about how to ascertain identity - maybe use two keys?

ranvirranaiitb:
> NB: There are only a few elliptic curves supported by this scheme, so if we want to switch to k-of-n signatures with partial signatures on votes, we may have to think about how to ascertain identity - maybe use two keys?

There may not be a need to ascertain identity if more than 2f+1 nodes sign the same prepare message. An honest node will sign a prepare message only after receiving it from a leader.  We know that at least f+1 nodes have signed the prepare QC, hence the block corresponding to that prepare QC must be constructed by the leader.

aakoshh:
True, I only meant ascertaining the identity to check that we're only talking to federation members. If the partial signature scheme was incompatible with the libp2p certificate extension then we'd need to think of how to bridge this gap.

ranvirranaiitb (approved):
We can add the other fields like `creator.signature`, `height`, `viewNumber`, `content merkleRoot` to the block header at a later stage.

PR #20, opened 2021-03-31
PM-2932: Checkpoint certificate data structure

aakoshh:
Adds the `CheckpointCertificate` type along with a `MerkleTree` component to hold the types for a Merkle root hash and a Merkle proof. Hashing/proving is not implemented in this PR.

Just realised that the certificate is supposed to contain block headers, so all models need to be visible to the interpreter, so I added a `checkpointing.models` module to hold everything together, referenced by both `checkpointing.service` and `checkpointing.interpreter`.

Also added a module reference to the `hotstuff.consensus` to be able to use the `QuorumCertificate` in the checkpoint.

### DIFF
--- a/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
+++ b/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
@@ -1,0 +1,66 @@
+package io.iohk.metronome.checkpointing.models
+
+import io.iohk.metronome.core.Tagger
+import scodec.bits.ByteVector
+
+object MerkleTree {
+  object Hash extends Tagger[ByteVector] {
+    val empty = apply(ByteVector.empty)
+  }
+  type Hash = Hash.Tagged
+
+  /** Merkle proof that some leaf content is part of the tree.
+    *
+    * It is expected that the root hash and the leaf itself is available to
+    * the verifier, so the proof only contains things the verifier doesn't
+    * know, which is the overall size of the tree and the position of the leaf
+    * among its siblings leaves. Based on that it is possible to use the sibling
+    * hash path to check whether they add up to the root hash.
+    *
+    * `leafCount` gives the height of the binary tree: `leafCount = 2^h`
+    * `leafIndex` can be interpreted as a binary number, which represents
+    * the path from the root of the tree down to the leaf, with the bits
+    * indicating whether to go left or right in each fork, while descending
+    * the levels.
+    *
+    * For example, take the following Merkle tree:
+    * ```
+    *       h0123
+    *     /       \
+    *   h01       h23
+    *  /   \     /   \
+    * h0    h1  h2    h3
+    * ```
+    *
+    * Say we want to prove that leaf 2 is part of the tree. The binary
+    * representation of 2 is `10`, which, we can interpret as: go right,
+    * then go left.
+    *
+    * The sibling path in the proof would be: `Vector(h3, h01)`.
+    *
+    * Based on this we can take the leaf value we know, reconstruct the hashes
+    * from the bottom to the top, and compare it agains the root hash we know:
+    *
+    * ```
+    * h2    = h(value)
+    * h23   = h(h2, path(0))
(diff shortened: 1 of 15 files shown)


ranvirranaiitb:
> Just realised that the certificate is supposed to contain block headers, so all models need to be visible to the interpreter, so I added a checkpointing.models module to hold everything together, referenced by both checkpointing.service and checkpointing.interpreter

I have a broad question:  Since there will be machines that participate in only ETC and no hotstuff, is it possible to structure the code in such a way that we can turn off checkpointing.service and Hotstuff without affecting the checkpointing.interpreter. I guess it is not important now, but might be useful later when we start writing actors.

ranvirranaiitb:
This PR looks great! 
I was initially worried that allowing a dynamic validator set will need a modification to the checkpoint certificate owing to some bribery attacks (past validators can be bribed to generate duplicate signatures). But Fisher and I concluded that such attacks can be easily handled by always following checkpoint certificate with the highest view number. 
Since the current version of the certificate carries a view number in commit QC, no modification is needed

ranvirranaiitb (approved):

aakoshh:
> This PR looks great!
Thanks! Do you implicitly approve its predecessor https://github.com/input-output-hk/metronome/pull/18 as well?

You're already thinking about how to pick between forks that have different certificates? Cool, we'll get to the nitty gritty soon :smiley: 

> I have a broad question: Since there will be machines that participate in only ETC and no hotstuff, is it possible to structure the code in such a way that we can turn off checkpointing.service and Hotstuff without affecting the checkpointing.interpreter. 

Definitely! My intention is that HotStuff and the Checkpointing Service live on once side of the fence, and the Checkpointing Interpreter on the other. I think the Interpreter can be the one that has generic classes for producing checkpoint candidates, runs validation, and understands (in a generic sense) what goes in the byte content of a transaction. These base classes would have hooks that the ETC or Advocate interpreter implements, but they could share a lot. 

You're asking about nodes that aren't federation members; I think most of what these need to do should be implemented in `checkpointing.models` actually, which is to validate the signatures of a Checkpoint Certificate, and then to unpack it. Most likely though the handling of the unpacked content would be in the ETC interpreter library; that's where we'd define what the contents of an ETC specific checkpoint candidate is. So it would be just a matter of configuration: if the node is not configured with a federation member private key, it doesn't start the Checkpointing Service.

> I guess it is not important now, but might be useful later when we start writing actors.

Hopefully we will not have actual `Actor`s in this one, just event processing loops, all statically typed and compiler checked.

ranvirranaiitb:
Thanks for the explanation, good to hear that we won't have `Actor` in this implementation. 

> Do you implicitly approve its predecessor #18 as well?

Sure, we can add the other fields `signature`, `height`, `viewNumber`, content `merkleRoot` at a later stage.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -0,0 +1,66 @@
...
+    * would conceptually be longer, e.g. `0010` for a tree with 16 leaves.
+    */
+  case class Proof(
+      // Position of the leaf in the lowest level.
+      leafIndex: Int,
+      // Number of leaves in the lowest level.
+      leafCount: Int,
+      // Hashes of the "other" side of the tree, level by level,
+      // starting from the lowest up to the highest.
+      siblingPath: IndexedSeq[MerkleTree.Hash]
```
aakoshh, line 64: @rtkaczyk suggested yesterday that we could get rid of `leafIndex` and `leafCount` from the `Proof` if instead we redefined the hash function from `def h(a, b) = hash(a ++ b)` to `def h(a, b) = hash(min(a, b), max(a, b))`, i.e. make the hash operation commutative. 

In that case we could still prove `Set` membership, but not the position among the transactions, so I suppose we could call it a `MerkleTreeSet`, but that would be enough for the proof for the BFT agreement, and if we keep the RLP based `bodyHash` in the `Block.Header` then it's also safe against reordering of transactions, in other contexts.

What do you think @ranvirranaiitb, @fishermanymc?

PR #22, opened 2021-04-08
PM-3104: Generic BlockStorage component for HotStuff

aakoshh:
Created a `BlockStorage` component under `hotstuff.service` that stores anything that has a `hotstuff.consensus.basic.Block` type class. The component manages parent-child relationships and makes available some traversal methods that were useful in tentative branch management. 

Also supports a pruning method which can be used to get rid of any block that isn't a descendant of a given hash. This can be used for example to periodically walk up the main chain from the last committed block, and get rid of anything that isn't a descendant of a block that is the, say, the 100th ancestor of the last committed block:

```scala
def prune[F[_]: Sync](
  db: RocksDBStore[F], 
  blockStorage: BlockStorage[RocksDB.Namespace, CheckpointingAggreement], 
  commitQC: QuorumCertificate[CheckpointingAgreement],
  maxChainLength: Int
): F[Unit] = {
  val program: KVStore[RocksDB.Namespace, Unit] = 
    for {
      mainChain <- blockStorage.getPathFromRoot(commitQC.blockHash).lift
      maybeNewRoot = mainChain.dropRight(maxChainLength).headOption
      _ <- maybeNewRoot match {
        case None => KVStore[RocksDB.Namespace].unit
        case Some(newRoot) => blockStorage.pruneNonDescendants(newRoot).void
      }
    } yield ()

  db.runWithBatching(program)
}
```

I thought about wrapping blocks in some meta-data to track timestamps, but I'm hoping that pruning can be done based on just the volume of information, i.e. we can always assume that everyone has synced the last 100 blocks.

### DIFF
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
@@ -28,6 +28,7 @@ class BlockStorage[N, A <: Agreement: Block](
 
     blockColl.put(blockHash, block) >>
       childToParentColl.put(blockHash, parentHash) >>
+      parentToChildrenColl.put(blockHash, Set.empty) >>
       parentToChildrenColl.update(parentHash, _ + blockHash)
   }
 
@@ -85,14 +86,15 @@ class BlockStorage[N, A <: Agreement: Block](
     * without any checking for the tree structure invariants.
     */
   private def deleteUnsafe(blockHash: A#Hash): KVStore[N, Unit] =
-    blockColl.delete(blockHash) >>
-      childToParentColl.get(blockHash).flatMap {
-        case None =>
-          KVStore[N].unit
-        case Some(parentHash) =>
-          childToParentColl.delete(blockHash) >>
-            parentToChildrenColl.update(parentHash, _ - blockHash)
-      }
+    childToParentColl.get(blockHash).flatMap {
+      case None =>
+        KVStore[N].unit
+      case Some(parentHash) =>
+        parentToChildrenColl.update(parentHash, _ - blockHash)
+    } >>
+      blockColl.delete(blockHash) >>
+      childToParentColl.delete(blockHash) >>
+      parentToChildrenColl.delete(blockHash)
 
   /** Get the ancestor chain of a block from the root,
     * including the block itself.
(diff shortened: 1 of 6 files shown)


metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
```diff
@@ -0,0 +1,190 @@
...
+    *
+    * This is true if the block has no children,
+    * or it has no parent and at most one child.
+    */
+  private def canDelete(blockHash: A#Hash): KVStoreRead[N, Boolean] =
+    (hasParent(blockHash), childCount(blockHash)).mapN {
+      case (_, 0)     => true
+      case (false, 1) => true
+      case _          => false
+    }
```
ranvirranaiitb, line 69: Should we also check if the block or its ancestors have not been committed? Or is that check performed somewhere else?

aakoshh: This one is a pure storage component, it doesn't have a concept of committed blocks. However, even if an ancestor is committed, it can be deleted if we want to free up space and our domain supports it, as it does in the case of Checkpointing: if we have a Q.C. for a checkpoint, we don't need the checkpoints before that, so we don't need to download the block history, and if we have it we can get rid of it after a while to keep storage space in check. 

Other applications may not have this property, in which case they wouldn't call pruning or deletion.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
```diff
@@ -0,0 +1,190 @@
...
+class BlockStorage[N, A <: Agreement: Block](
+    blockColl: KVCollection[N, A#Hash, A#Block],
+    childToParentColl: KVCollection[N, A#Hash, A#Hash],
+    parentToChildrenColl: KVCollection[N, A#Hash, Set[A#Hash]]
+) {
+  private implicit val kvn  = KVStore.instance[N]
+  private implicit val kvrn = KVStoreRead.instance[N]
+
+  /** Insert a block into the store, and if the parent still exists,
+    * then add this block to its children.
```
rtkaczyk, line 24: `canDelete` performs checks that disallow turning the tree into the forest. But this method allows us to put an orphaned block. Should we perform a check for parent here? Or is a forest allowed, and if so, in what situations would that be useful?

aakoshh: Good observation! That is correct, it doesn't prevent the caller from inserting a forest. 

As the main docstring says, the store assumes that the the caller stores at least one pointer to the tree. If you insert independent nodes and store pointers to it, that's fine as you can still traverse everything you inserted. Maybe you wanted to store two trees, like the Prism voter chains, they each have a different genesis block.

What the store is trying to avoid is that we accidentally delete something and make parts of the tree untraversable, so the caller should trim the trees from the edges. 

I mainly wanted to avoid leaking space during pruning due to some mistake. 

In theory the store could store a pointer to the root, and disallow insertions that don't have a parent after a root has been established, but I thought since the quorum certificates store block hashes anyway, it would just slow things down and be more complicated.

rtkaczyk (approved):

PR #23, opened 2021-04-09
PM-2936: Remote messaging protocol

aakoshh:
Adds message types for synchronisation between nodes.

I tried different approaches:
1. There was a `ProtocolMessage` in `hotstuff.service` and a `RemoteMessage` type in `checkpoint.service` that multiplexed between BFT and application specific messages. However I didn't quite like it because I wanted to keep as much of the generic application structure as possible in the `hotstuff.service` module, so that swapping out the `checkpointing` use case to something else reuses as much as possible. Having the multiplexer closer to `hotstuff` seemed closer to that goal.
2. There's now a `HotStuffMessage` and a `DuplexMessage` in `hotstuff.service`, the latter being generic in what kind of application message we need to support. If we don't need one, we can just use `HotStuffMessage`. These types could be replaced by aliases to `Either`.

The idea is that we'll instantiate our network component like so:
```scala
type RemoteMessage = DuplexMessage[CheckpointingAgreement, CheckpointingMessage]
val network = RemoteConnectionManager[RemoteMessage](...)
```
But in a different use case it could be:
```scala
type RemoteMessage = HotStuffMessage[MyAgreement]
val network = RemoteConnectionManager[RemoteMessage](...)
```
And somewhere will be a dispatcher that handles `RemoteMessage`, where the compiler will make us deal with all the possible message types, whether it's wrapped in a duplex format or not.


metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/Status.scala
```diff
@@ -0,0 +1,16 @@
...
+
+/** Status has all the fields necessary for nodes to sync with each other.
+  *
+  * This is to facilitate nodes rejoining the network,
+  * or re-syncing their views after some network glitch.
+  */
+case class Status[A <: Agreement](
+    viewNumber: ViewNumber,
+    prepareQC: QuorumCertificate[A],
+    commitQC: QuorumCertificate[A]
```
ranvirranaiitb, line 15: Might be a good idea to add `lockedQC`. Although it is difficult to prove that a node is locked to that QC, it might be helpful to handle dynamic validator sets.

aakoshh: We can add anything, but I thought we can do so when we know how we'll use them in the synchronisation process, perhaps when we have a spec for the handing of dynamic validator sets? I thought these two are the minimum that is required: the highest `prepareQC` is going to be the basis for the next round, and the `commitQC` is where we can sync from. Not sure how to use `lockedQC` right now.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/messages/CheckpointingMessage.scala
```diff
@@ -0,0 +1,31 @@
+package io.iohk.metronome.checkpointing.service.messages
+
+import io.iohk.metronome.checkpointing.models.Ledger
+import io.iohk.metronome.core.messages.{RPCMessage, RPCMessageCompanion}
+
+/** Checkpointing specific messages that the HotStuff service doesn't handle,
+  * which is the synchronisation of committed ledger state.
+  *
+  * These will be wrapped in an `ApplicationMessage`.
```
rtkaczyk, line 9: Why so? I would expect the networking and RPC layers to be unrelated, and so their messages too.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/messages/DuplexMessage.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import io.iohk.metronome.hotstuff
+import io.iohk.metronome.hotstuff.consensus.basic.Agreement
+
+/** Messages type to use in the networking layer if the use case has
+  * application specific message types, e.g. ledger synchronisation,
+  * not just the general BFT agreement (which could be enough if
+  * we need to execute all blocks to synchronise state).
+  */
+sealed trait DuplexMessage[A <: Agreement, M]
```
rtkaczyk, line 11: Do we need the 2 type params?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/messages/DuplexMessage.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object DuplexMessage {
+
+  /** General BFT agreement message. */
+  case class AgreementMessage[A <: Agreement](
+      message: hotstuff.service.messages.HotStuffMessage[A]
```
rtkaczyk, line 17: Do we need a qualified name here?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/messages/CheckpointingMessage.scala
aakoshh: The `networking` module is indeed unrelated to the RPC layer, in the sense that it doesn't know what kind of messages it is transmitting, it's [generic in the message type](https://github.com/input-output-hk/metronome/blob/develop/metronome/networking/src/io/iohk/metronome/networking/RemoteConnectionManager.scala#L21). 

But in the RPL layer I see our application as having two components:
* generic HotStuff consensus, which can be applied to any `Agreement`
* agreement specific extras, which in our case is Checkpointing which can directly exchange the Ledger state, but it could be something else that may require its own different set of messages.

So I want the `RemoteConnectionManager` to be able to handle the union of these messages, hence the wrapping, and we'll have something like a `RemoteConnectionManager[Task, CheckpointingAgreement.PKey, Either[HotStuffMessage, CheckpointingMessage]`. I just made that explicit by having a `DuplexMessage` instead of the `Either`. 

Does that answer your question? `ApplictionMessage` has nothing to do with the network, it's just a wrapper that I came up with so that we can compose a network that has application side messages, _if_ if needs them.

So our network - RPC stack would look like this:
```
              -----------
              | Network |
              -----------
                  |
           -----------------
           |               |
     ------------     -----------------
     | HotStuff |     | Checkpointing |
     ------------     -----------------
           |
    -------------------
    |                 |
-------------      --------
| Consensus |      | Sync |
-------------      --------
```
With some further pipes between components that need to send messages to each other as well.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/messages/DuplexMessage.scala
aakoshh: See my [other comment](https://github.com/input-output-hk/metronome/pull/23#discussion_r614724939) about what I think HotStuff setups would look like: they have the generic and the application specific part where we can dispatch messages to.

I think not all applications need their specific messages, though. For example I'd like to do an example application which just agrees on the order of some "commands", e.g. randomly generated strings. These wouldn't have a ledger for example, just a block history, like in the HotStuff paper, so they don't need the equivalent of the `CheckpointingMessage`. I could have moved this `DuplexMessage` into the `checkpointing` module, in which case it wouldn't need a generic parameter, however I thought that the `hotstuff` module should capture as much of the reusable patterns as possible, so the application specific part can be thin. 

This could be replaced with just an `Either`, but I thought it serves as better documentation of what a reusable pattern could be, in that you either instantiate a `RemoteConnectionManager[Task, PublicKey, HotStuffMessage[YourAgreement]]` or a `RemoteConnectionManager[Task, PublicKey, DuplexMessage[YourAgreement, YourMessages]]` and we can provide RLP formatters for the `DuplexMessage`, rather than all `Either` in general.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/messages/DuplexMessage.scala
aakoshh: Not really, just left it for consistency with what the `HotStuffMessage` itself looks like, where it is needed for `SyncMessage` but increases the readability for the `Message` as well. Can remove it if you prefer.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/messages/CheckpointingMessage.scala
aakoshh: The HotStuff side of that diagram is reflected in the [HotStuffService](https://github.com/input-output-hk/metronome/pull/25/files#diff-0f291caa9674678c06eec2f7fe2c17c89cb58bf117c8cf5a67ecd8c03d3360bb)

rtkaczyk: I may be a bit confused here. Is `CheckpointMessage` meant be exchanged between Metronome nodes or between a Metronome node and an associated PoW node?

aakoshh: I'm sorry, I found it difficult to come up with names; I named it after the eventual `CheckpointingService`, so `ConsensusMessage` goes to the `ConsensusService`, `SyncMessage` goes to the `SyncService`. 

The `CheckpointingMessage` is stuff that goes between Metronome nodes, that's why it's under the "Remote messaging protocol" PR. 

The proposal for messages that go between the Metronome node and the associated PoW node are in https://github.com/input-output-hk/metronome/pull/24

aakoshh: The `CheckpointingMessage` is what allows two nodes to synchronise their ledger state; it could work in different ways, though, like synchronising a Merkle Patricia Trie, for just the parts that changed. It's just that in our case we know that our `Ledger` resets with every checkpoint, so it's small enough to be exchanged in a single message.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/messages/DuplexMessage.scala
rtkaczyk: What I don't get is how this union type is better than simple subclassing of a common `Message` class, especially that `M` is unbounded.

rtkaczyk: As I understand there won't be a case where both `A` and `M` are defined as something other than `Nothing`, would they?

aakoshh: What should subclass a common message class exactly, and where would the baseclass live?

Say I don't need application specific messages, I just want to use the `HotStuffMessage`. Should it be a sublcass of some higher level message type? What purpose would that serve in that scenario?

On the other hand, if I want to be delivered my application messages to my application service, again why should it inherit some class from HotStuff? It's the only kind that should arrive there. 

So, nothing should ever have to deal with the base class.

> there won't be a case where both A and M are defined as something other than Nothing, would they?

Sorry I don't understand, the case for checkpointing will be that both are not `Nothing`, `A` going to be the `CheckpointingAgreement` and `M` the `CheckpointingMessage`. I'm probably misunderstanding the question.

rtkaczyk: Ok, let me put it another way. You mention that `DuplexMessage` could replaced by `Either`. So what additional decision is there to be made (and where) based on whether it's a `Left` or `Right`? Why can't that decision be made based on the message type alone?

aakoshh: The decision that has to be made based on whether we're dealing with a `Left` or `Right` is whether an incoming network packet has to be delivered to the HotStuff components or the Checkpointing ones. 

There were other options: 
* First I thought I'll just add a generic `ApplicationMessage` to the `HotStuffMessage`, but the problem is that it's not necessary to have one.
* Or I thought the Checkpointing side could be in charge of routing to the HotStuff side, since it knows about them (the module dependencies allow that in that direction), but I wanted to keep the Checkpointing side thin, so if we had another variant there's nothing to copy paste.
* So I ended up with this: we have `Either[Consensus, Sync]` or we have `Either[[Either[Consensus, Sync], Application]`. The HotStuff module can support both with these types, allowing the application root to assemble the correct stack.

> Why can't that decision be made based on the message type alone?

We have to instantiate the network stack with some kind of message type that has a `Codec`. If I used a single `Message` type, perhaps an unsealed trait that is extended in the checkpointing module, I think I'd lose the help from the compiler to tell me if I forgot to implement something.

rtkaczyk: Ok, I guess the need for the `Codec` justifies this design, but I still don't understand why the `DuplexMessage` has to carry the `Agreement` type. I doesn't look like `RemoteConnectionManager` knows anything about `Agreements`.

Also:
> the case for checkpointing will be that both are not Nothing, A going to be the CheckpointingAgreement and M the CheckpointingMessage

but the docstring says: "These will be wrapped in an `ApplicationMessage`" and `ApplicationMessage` is defined as `DuplexMessage[Nothing, M]`

aakoshh: The `DuplexMessage` carries the agreement type because it's part of the HotStuff module and one of its cases is fixed to be a `HotStuffMessage`; it's sole purpose is to combine `HotStuffMessage` with some application specific message family, not to be a replacement of `Either` where both sides can change freely. The `RemoteConnectionManager` doesn't know anything about `Agreement` but it also doesn't know anything about `DuplexMessage`. I though `ExtendedHotStuffMessage` could be a name as well, or something similar, but landed on "duplex" as in a duplex house with two sides.

About `Nothing`: now I understand what you meant. I'll probably have to add a `+` before `M` and `A` at some point, but always use `Nothing` when that generic parameter doesn't appear in the case class itself, for example [here](https://github.com/input-output-hk/metronome/blob/develop/metronome/networking/src/io/iohk/metronome/networking/NetworkEvent.scala#L43-L48). Otherwise I'd always have to specify what the other type is when constructing, say, an `AgreementMessage`. This seems to be [common practice](https://github.com/fpinscala/fpinscala/blob/master/answers/src/main/scala/fpinscala/errorhandling/Either.scala#L27-L28). 

I thought you meant whether we'd create a `RemoteConnectionManager[PublicKey, DuplexMessage[SomeAgreement, Nothing]]`.

rtkaczyk: Sure, I'm not questioning the practices. However, I'm still not clear on the type choice. Maybe because currently I can only see `DuplexMessage` as being passed to `RemoteConnectionManager` and it will become more clear when I see the messages being routed and interpreted. Sorry for being so inquisitive, I don't mean to hinder the progress here, just to build a solid understanding of the code.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/messages/HotStuffMessage.scala
```diff
@@ -0,0 +1,23 @@
...
+import io.iohk.metronome.hotstuff.consensus.basic.Agreement
+
+/** Messages which are generic to any HotStuff BFT agreement. */
+sealed trait HotStuffMessage[A <: Agreement]
+
+object HotStuffMessage {
+
+  /** Messages which are part of the basic HotStuff BFT algorithm itself. */
+  case class ConsensusMessage[A <: Agreement](
+      message: hotstuff.consensus.basic.Message[A]
```
rtkaczyk, line 13: I think it might be more intuitive to browse the project structure if this `Message` also lived in its own `messages` sub-package, symmetric to other modules

aakoshh: You mean inside `consensus.basic.messages`? Yeah maybe, it could also perhaps be called `ProtocolMessage` to help recognise what it is. At the time I thought a flat package is easy enough to understand, everything is just about the `basic` protocol. 

I'd leave it there for now so there's less churn on this chain of PRs, but we can move it around if it makes it clearer.

rtkaczyk (approved):

PR #24, opened 2021-04-10
PM-2937: Local messaging protocol

aakoshh:
Added a `checkpointing.interpreter.messages.InterpreterMessage` to capture the requests and responses that can go between the Checkpointing Service and the Checkpointing Interpreter.


metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/messages/InterpreterMessage.scala
```diff
@@ -0,0 +1,147 @@
...
+    *
+    * If the block was valid, `Some` updated ledger is returned,
+    * otherwise the response contains `None`.
+    *
+    * The Service will check whether the hash of the updated ledger
+    * matches the `postStateHash` in the block header.
+    */
+  case class ValidateBlockResponse(
+      requestId: RequestId,
+      maybeUpdatedLedger: Option[Ledger]
```
aakoshh, line 119: Here we could use a `result: Either[Error, Ledger]` if there is something to be communicated back about the reason for failure.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/messages/InterpreterMessage.scala
```diff
@@ -0,0 +1,148 @@
...
+  /** The Interpreter responds to the block validation request when
+    * it has all the data available to perform the validation.
+    *
+    * The result indicates whether the block contents were valid.
+    *
+    * Reasons for being invalid could be that a checkpoint
+    * was proposed which is inconsistent with the current ledger,
+    * or that a proposer block was pointing at an invalid block.
+    *
+    * If valid, the Service updates its copy of the ledger
```
aakoshh, line 123: I had another design where `ValidateBlockResponse` contained the updated ledger, however I realised that it wouldn't be symmetric with how the ledger is updated during block creation, which, by the shape of these messages, would happen in the Service. 

Another design option would be to delegate the block (body) creation on the Interpreter side, which is I think what we discussed with @ranvirranaiitb. That would looks something like this:

```scala
  case class CreateBlockRequest(
      requestId: RequestId,
      ledger: Ledger
  ) extends InterpreterMessage
      with Request
      with FromService

  case class CreateBlockResponse(
      requestId: RequestId,
      maybeResult: Option[CreateBodyResponse.Result]
  ) extends InterpreterMessage
      with Response
      with FromInterpreter
  object CreateBlockResponse {
    case class Result(
      blockBody: Block.Body, 
      updatedLedger: Ledger
    )
  }

  case class ValidateBlockRequest(
      requestId: RequestId,
      blockBody: Block.Body,
      ledger: Ledger
  ) extends InterpreterMessage
      with Request
      with FromService

  case class ValidateBlockResponse(
      requestId: RequestId,
      maybeUpdatedLedger: Option[Ledger]
  ) extends InterpreterMessage
      with Response
      with FromInterpreter {
    def isValid = maybeUpdatedLedger.isDefined
  }

  case class NewCheckpointCertificateRequest(
      requestId: RequestId,
      checkpointCertificate: CheckpointCertificate
  ) extends InterpreterMessage
      with Request
      with FromService
      with NoResponse
```

In this version the Interpreter would be solely responsible for updating the ledger and wouldn't notify the Service about individual proposer events, it would just be asked when it's leading to put whatever it has into a new block body, based on whatever the current ledger state is.

The difficulty there is how to manage the mempool. The Service wouldn't be able to maintain one, because it's never notified about proposer events, and the Interpreter doesn't have access to blocks, so it can't just filter whatever it has in memory by the branch the block being proposed is on. 

Yet another option is to keep the proposer notifications to maintain the mempool, and pass the mempool back as part of `CreateBlockRequest`, and its updated value in `CreateBlockResponse` and `ValidateBlockResponse`.

I think all these can be avoided if we can keep the ledger update rules transparent to the system. I asked this before, but I think the only unclear scenario would be this:
```
<- NewProposerBlockRequest(block99)
<- NewProposerBlockRequest(block100)
<- NewCheckpointCandidateRequest
<- NewProposerBlockRequest(block101)
-> GetCheckpointCandidateRequest(ledger = List(block99, block100, block101))
<- GetCheckpointCandidateResponse(checkpoint = block100)
```
In that sequence of events, the checkpoint should only apply up to block 100, block 101 is not checkpointed, so the updated ledger should be `Ledger(checkpoint = block100, proposerBlocks = List(block101)`. What we should avoid doing is to append all proposer events to the ledger, then clear it out with an (opaque) checkpoint that, unbeknown to us, doesn't have block101, so we just lost an event.

For that to happen, the Service would have to remember the position of the checkpoint notification among the proposer blocks, and during block creation apply the proposer blocks before the notification, then ask for the candidate, then apply the rest:
```scala
val ledger1 = ledger0.update(block99).update(block100)
val checkpoint = interpeter.getCheckpointCandidateRequest(ledger1)
val ledger2 = ledger1.update(checkpoint)
val ledger3 = ledger2.update(block101)
val block = Block(..., ledger3.hash, Body(block99, block100, checkpoint, block101))
```

Because we're using Advocate, we're assuming that a checkpoint will eventually contain all proposer blocks, so for an honest node, the mempool can be filtered by the proposer events it sees in the block bodies, there's no other hidden logic there. As long as it only accepts proposer events from its (honest) PoW Interpreter, the Service shouldn't have a value in its mempool that will never clear out, or is invalid according to the other federation members. They also shouldn't have a checkpoint that removes it, without a corresponding proposer event in a block.

It would be possible, though, to receive a block which has block99, block100 and block101 in its body without a checkpoint, however, as @ranvirranaiitb said, the Interpreter should reject this block as invalid. Otherwise, in the next step, it would propose a checkpoint which doesn't cover block101, but the Service wouldn't know this and clear it out from the ledger. 

Which way do you think we should go?

rtkaczyk: What if the Interpreter after completing the `GetCheckpointCandidateResponse` would re-send the proposer blocks that were not covered by the checkpoint? (We could then just blindly clear out the mempool)

rtkaczyk: Sounds good

rtkaczyk (approved):
Looks pretty clear :+1:
We may want to update the [message sequence diagram] though, I find it quite helpful. Could be done after this becomes stable.

aakoshh:
Yes, a full specification for message passing is one of the first items @fishermanymc and the team are working on.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/messages/InterpreterMessage.scala
aakoshh: In retrospect I'm not sure why I wanted the include the `updatedLedger` in the response. If the `CreateBlockResponse` has the block body, it should be enough to ensure that updating the block body to the ledger is something the Service can do on its own, and the Interpreter could make sure that all proposer events are ordered after the checkpoint, among the transactions in the body. 

Maybe passing the filtered mempool along the request is not a bad idea, so the service can present the transactions that are currently in it after all transactions in pending blocks have been tentatively removed from it:

```scala
  case class CreateBlockRequest(
      requestId: RequestId,
      ledger: Ledger,
      mempool: Seq[Transaction.ProposerBlock]
  ) extends InterpreterMessage
      with Request
      with FromService

  case class CreateBlockResponse(
      requestId: RequestId,
      maybeBlockBody: Option[Block.Body]
  ) extends InterpreterMessage
      with Response
      with FromInterpreter
```

aakoshh: I had this example for the case where we'd pass the mempool along as part of the create block request:

0. Say we want to checkpoint every 5th block and the current ledger is `{last_checkpoint: 5, proposer_blocks: [6,7]}`. These proposers have been validated by everyone.
0. Alice's mempool has `[8a, 9a]`
0. Bob's mempool has `[8b, 9b, 10b, 11b]`
0. Bob is the leader, his service calls: `interpreter.createBlock(ledger = {last_checkpoint: 5, proposer_blocks: [6,7]}, mempool = [8b, 9b, 10b, 11b])`
0. Bob's interpreter looks at the PoW chain and realises that `10b` is a block to be checkpointed; it returns `Body(transactions = [8b, 9b, 10b, CP(10b), 11b])`
0. Bob proposes that block to the federation.
0. Alice can't vote on immediately, she doesn't have some of the blocks. But let's say the others do.
0. Bob, and the other execute the block.
0. They apply the body on the ledger, which gives `{last_checkpoint = 10b, proposer_blocks = [11b]}`. The appearance of `CP(10b)` clears out the proposer blocks that were in the ledger before it. If `11b` appeared before `CP(10b)` then honest federation members would have rejected the block because they'd see that it didn't insert a CP at a height it should have.
0. They apply the block on their respective mempools. Bob's mempool becomes empty because everything he had appeared in the body. Alice's mempool still has` [8a, 9a]`, which is another fork and didn't appear in the body.
0. Alice is leader, her service calls `interpeter.createBlock(ledger = {last_checkpoint = 10b, proposer_blocks = [11b]}, mempool = [8a, 9a])`
0. Alice's interpreter would be using Advocate and know that since these are honest blocks, she can enforce their inclusion in the next checkpoint still, so it would return `Body(transactions = [8a, 9a])`
0. When her block gets committed, she can remove the blocks from her mempool.

So if we use Advocate, we don't actually have to worry about clearing the mempool, because eventually everything has to appear in a block. Otherwise there would perhaps have to be a further mechanism, e.g. filter the block as part of the block execution.

aakoshh: Updated the code to reflect this version :point_up:

rtkaczyk (approved):
Still LGTM. `CreateBlockBodyRequest` makes more sense

PR #25, opened 2021-04-12
PM-3146: HotStuff Service

aakoshh:


rtkaczyk (approved):
Looks good, but I'm not yet quite clear on all the cats-effect mechanics used here

PR #26, opened 2021-04-14
[PM-3102,3103] MerkleTree construction and proof verification

rtkaczyk:
~Need to address a few TODOs~


metronome/checkpointing/models/test/src/io/iohk/metronome/checkpointing/models/RLPCodecsProps.scala
```diff
@@ -24,5 +24,5 @@ object RLPCodecsProps extends Properties("RLPCodecs") {
   propRoundTrip[Transaction]
   propRoundTrip[Block]
   propRoundTrip[ECDSASignature]
-  propRoundTrip[CheckpointCertificate]
+  //propRoundTrip[CheckpointCertificate]
```
aakoshh, line 27: Does it not work?

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -58,9 +61,124 @@ object MerkleTree {
...
+    *
+    * 1 - traverse right (take left hash)
+    */
+  private def navigateTree(height: Int, leafIndex: Int): Int =
+    leafIndex >> (height - 1) & 1
 }
+
+//TODO: seperate into `Node` and `Leaf`. This structure does not enforce integrity (left cannot be empty if
+// right is defined). Also doesn't handle empty trees very well.
+case class MerkleTree(
```
aakoshh, line 180: To enforce integrity, you could turn it into a `sealed abstract case class MerkleTree private(`. 
In that case the compiler wouldn't generate a `copy` and an `apply` method, and you'd have to produce instances smart constructors calling `new MerkleTree(...) {}`, but you could be sure that the integrity is preserved.

metronome/checkpointing/models/test/src/io/iohk/metronome/checkpointing/models/RLPCodecsProps.scala
rtkaczyk: No no, forgot to re-enable it

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -58,9 +61,124 @@ object MerkleTree {
...
+      }
+    }
+
+    verify(leaf, 1, proof.siblingPath) == root
+  }
+
+  def generateProof(root: MerkleTree, elem: Hash): Option[Proof] = {
+    val leaves    = findLeaves(root)
+    val leafIndex = leaves.indexOf(elem)
+    val height    = leafCountToHeight(leaves.size)
```
aakoshh, line 125: I was thinking that the concatenation of all leaves could be avoided by doing two things:
1. Store the size of the underlying Merkle tree along with the hash
2. Returning the index of the hash in the tree by traversal until its found and figuring out the index based on the path taken (building an index the opposite way as `navigateTree`).

It depends on the tree size I suppose and how many proofs we need to generate. In our case we'd only have to generate the proof for certain blocks, so we wouldn't have to store the trees where the sizes would take up space, and they are small anyway. 

Removing the `leafCount` from the `Proof` would make it even clearer that there's no need to concatenate the whole tree, since we don't need the full size, so let's do it!

Another observation is that if I know the hash of the item then then I probably know its index in the original tree as well: it's the position of the transaction in the body, so the index could be an input to the function. It could be the input _instead_ of the hash, so we don't have to re-hash the item, since the hash is available in the tree.

Maybe both are valid cases, if retrieving the block body is more costly and we already have the transaction at hand, ready to be hashed.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -58,9 +61,124 @@ object MerkleTree {
...
+  ): IndexedSeq[Hash] = tree match {
+    case MerkleTree(_, None, None) =>
+      IndexedSeq.empty
+
+    case MerkleTree(_, left, right) =>
+      val dir                 = navigateTree(height, leafIndex)
+      val (traverse, sibling) = if (dir == 0) (left, right) else (right, left)
+      val siblingHash         = sibling.map(_.hash).getOrElse(traverse.get.hash)
+
+      findSiblings(traverse.get, height - 1, leafIndex) :+ siblingHash
```
aakoshh, line 156: It would be nicer if you didn't append to a list, in general, even if our trees are small, it's not good practice.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -58,9 +61,124 @@ object MerkleTree {
...
+
+  /** `height = ceil(log2(leafCount))`
+    * Empty tree has height == 0
+    */
+  private def leafCountToHeight(leafCount: Int): Int = {
+    import math._
+    ceil(log(leafCount) / log(2)).toInt
+  }
+
+  //TODO: use an ADT for return instead?
```
aakoshh, line 167: Yeah, perhaps `dir == Left` would read better than `dir == 0`, or maybe `navigateTree` could be renamed to read something like `val goLeft = isLeftTurn(height, leafIndex)`.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -58,9 +61,124 @@ object MerkleTree {
       // Position of the leaf in the lowest level.
       leafIndex: Int,
       // Number of leaves in the lowest level.
+      // TODO: remove? not used when verifying the proof
```
aakoshh, line 64: Very elegant way of doing it :+1: 

I don't know if it helps people to see this height or not, but I'd say if we don't need it then we can remove it. In the three examples in our codebase 3 out of 2 have this number, but the explanation in the docstrings works without it as well.

aakoshh (commented):
Really nice, so much easier to read than the other examples I've seen, good stuff!

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
aakoshh: Would you prefer an ADT of `FullNode`, `LeftNode` (left child only) and `Leaf`?

aakoshh:
Would you be so kind as to implement [Merkleizing the in the Block](https://github.com/input-output-hk/metronome/blob/develop/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Block.scala#L46-L47) in this PR as well?

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
aakoshh: It looks like the size is used for sanity checking the sibling path and the index itself.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -58,9 +61,124 @@ object MerkleTree {
...
+    def buildTree(nodes: Seq[MerkleTree]): Seq[MerkleTree] = {
+      if (nodes.size <= 1)
+        nodes
+      else {
+        val paired = nodes.grouped(2).toSeq.map {
+          case Seq(a, b) =>
+            MerkleTree(hashFn(a.hash, b.hash), Some(a), Some(b))
+          case Seq(a) =>
+            // if the element has no pair we hash it with itself
+            MerkleTree(hashFn(a.hash, a.hash), Some(a))
```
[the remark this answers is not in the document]
rtkaczyk: I think all those options are viable. I initially just hashed the hash, but I ended up with self-concat thinking it would be least controversial, as that's what bitcoin does. The only thing I can think of is that the approaches you mention do reveal a bit more about the tree structure in the proof. Can't see how that could lead to any vulnerabilities.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
rtkaczyk: If it's a `Vector` then append and prepend complexity should be the same. I actually can't find info on whether `IndexedSeq` still defaults to `Vector` in 2.13, but I could make that explicit.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
rtkaczyk: You could also deduce the height from `siblingPath`

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
aakoshh: Ah, sorry I thought it was a `Seq` :+1:

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
aakoshh: I think it's fine to do the hash check and forego the sanity checking.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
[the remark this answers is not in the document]
aakoshh: Thanks, didn't know this was how bitcoin did it; that's good enough for me :)

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
rtkaczyk: So instead of `findLeaves` I ended up with functions `findHeight` and `findSize` functions which only traverse a single branch, `findElem` which return a single Int. I wouldn't expect our trees to be large enough to require such optimisation, but it was a fun exercise :)

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -1,14 +1,30 @@
...
+sealed trait MerkleTree {
+  def hash: MerkleTree.Hash
+}
+
 object MerkleTree {
   object Hash extends Tagger[ByteVector] {
-    val empty = apply(ByteVector.empty)
+
+    /** Indicates an empty MerkleTree */
+    val empty = apply(Keccak256(ByteVector.empty))
```
rtkaczyk, line 17: A request for comments: this was initially just an empty `ByteVector` (not hashed), but e.g. Ethereum does choose a 32-byte value for the case of empty set. I guess the question boils down to do we need this value to be of constant size?

aakoshh: The way for example Dmitry modeled the `Hash` type is that it asserts that it is of 32 bytes long, so I think an all zero hash, or this hash of an empty vector, would be better than an empty hash.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Block.scala
```diff
@@ -55,7 +55,6 @@ object Block {
     val header = Header(
       parentHash = Block.Header.Hash(ByteVector.empty),
       postStateHash = Ledger.empty.hash,
-      bodyHash = body.hash,
       contentMerkleRoot = MerkleTree.Hash.empty
```
aakoshh, line 58: Maybe the above is a better pattern, i.e. instead of `MerkleTree.Hash` having an `empty` value, it could be `MerkleTree.empty.hash`?

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -51,16 +66,148 @@ object MerkleTree {
...
+  /** Finds the tree size (number of leaves), by traversing the rightmost branch */
+  private def findSize(tree: MerkleTree, size: Int = 0): Int = tree match {
+    case Leaf(Hash.empty) =>
+      0
+    case Leaf(_) =>
+      size + 1
+    case Node(_, left, None) =>
+      findSize(left, size << 1)
+    case Node(_, _, Some(right)) =>
+      findSize(right, size << 1 | 1)
```
aakoshh, line 160: Nice!

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Transaction.scala
```diff
@@ -13,9 +14,13 @@ import scodec.bits.BitVector
...
   * and to register checkpoint candidates.
   */
-sealed trait Transaction
+sealed trait Transaction {
+  def hash: Transaction.Hash
+}
 
 object Transaction {
+  object Hash extends Tagger[ByteVector]
+  type Hash = Hash.Tagged
```
aakoshh, line 23: This should be enough, and then you don't need the overrides below:
```scala
sealed trait Transaction extends RLPHash[Transaction, Transaction.Hash]

object Transaction
    extends RLPHashCompanion[Transaction]()(RLPCodecs.rlpTransaction) {
```

It will include the transaction type tag in the RLP representation of the transaction instances.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
aakoshh: Really cool :)

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Block.scala
rtkaczyk: Yes, I like that

aakoshh (approved):
Good job, love our new tree :evergreen_tree: 

The only thing left to do is to add yourself to the devs in `build.sc` :partying_face:

aakoshh:
Looks like you haven't configured your git repo to sign the commits with gpg.

rtkaczyk:
Will squash and sign

PR #27, opened 2021-04-15
PM-3146: HotStuff Service Tests

aakoshh:
Adding tests for some of the components added in https://github.com/input-output-hk/metronome/pull/25

### DIFF
--- a/metronome/core/test/src/io/iohk/metronome/core/fibers/FiberSetSpec.scala
+++ b/metronome/core/test/src/io/iohk/metronome/core/fibers/FiberSetSpec.scala
@@ -1,0 +1,83 @@
+package io.iohk.metronome.core.fibers
+
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import monix.execution.atomic.AtomicInt
+import org.scalatest.compatible.Assertion
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.Inside
+import scala.concurrent.duration._
+
+class FiberSetSpec extends AsyncFlatSpec with Matchers with Inside {
+
+  def test(t: Task[Assertion]) =
+    t.timeout(10.seconds).runToFuture
+
+  behavior of "FiberSet"
+
+  it should "reject new submissions after shutdown" in test {
+    FiberSet[Task].allocated.flatMap { case (fiberSet, release) =>
+      for {
+        _ <- fiberSet.submit(Task("foo"))
+        _ <- release
+        r <- fiberSet.submit(Task("bar")).attempt
+      } yield {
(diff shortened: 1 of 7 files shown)


metronome/core/test/src/io/iohk/metronome/core/fibers/FiberSetSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  it should "process tasks concurrently" in test {
+    FiberSet[Task].use { fiberSet =>
+      val running    = AtomicInt(0)
+      val maxRunning = AtomicInt(0)
+
+      for {
+        handles <- Task.traverse(List.fill(10)(())) { _ =>
```
rtkaczyk, line 66: Minor: `Task.traverse(1 to 10)` would be nicer to read

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/ConsensusServiceSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+  ProtocolError,
+  Event,
+  Message,
+  Phase,
+  QuorumCertificate
+}
+import io.iohk.metronome.hotstuff.consensus.ViewNumber
+import io.iohk.metronome.crypto.GroupSignature
+
+class ConsensusServiceSpec extends AsyncFlatSpec with Matchers {
```
rtkaczyk, line 20: Are we going to add more tests here once we fill in the missing bits in `ConsensusService`? Regardless, since this tests only the `MessageStash` maybe it's better have be just about that and rename to `MessageStashSpec`?

metronome/core/test/src/io/iohk/metronome/core/fibers/FiberSetSpec.scala
aakoshh: Yeah, doesn't look nice for sure :+1:

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/ConsensusServiceSpec.scala
aakoshh: I haven't added more tests yet. Although I feel a little bad about that, it looked like it's more trouble than its worth, and wanted to just do an integration test that somehow covers the whole system behaviour, or to see how it can be refactored into smaller components like this `MessageStash` that are easy to test.

I thought since it's a nested class it can be inside the `ConsensusServiceSpec` and this `behave like` makes it easy to see what belongs to it, the code can be collapsed as one region as well. 

But if you think it's better to have it in a separate class, that's fine with me.

rtkaczyk (approved):
Nice, concise tests

PR #29, opened 2021-04-17
PM-3104: Insert blocks in any order

aakoshh:
Changes the `BlockStore` in anticipation of block synchronisation to be able to insert blocks in reverse, or any other order, and still maintain parent-child relationships. 

Currently if a child is inserted before its parent, then the parent-to-children association will not be populated, and traversal from ancestors towards descendants will not work. After the change, the parent-to-children association is established by the insertion of the children. The removal happens when the last child _and_ the parent are all removed. 

Added a `KVStore.alter` method (like `Data.Map.alter` in Haskell) to be able to insert/update/delete in one operation.

I expect that during block synchronisation we'll start walking the tree backwards until one of these conditions hold:
* we find a block we already have in storage
* we find a Commit Q.C. in which case we don't need the ancestry, so we can clear our block storage and re-insert from the new root

For the first case we'll insert blocks going from child to parent, and this change will ensure that during that we can still traverse the stored blocks in both ways. This is true for in-memory as well as persistent storage. 

NOTE: The `parentToChildrenColl` will no longer contain an empty set when a block is inserted, only when its first child is inserted. The reason for this was so that when we delete the last child in `deleteUnsafe` we can safely remove an empty Set; otherwise we'd have to check that the parent doesn't exist, which would incur yet another query. The only repercussion was that `getDescendants` has to check whether a single result is due to non-existent block or empty children.

### DIFF
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
@@ -29,9 +29,6 @@ class BlockStorage[N, A <: Agreement: Block](
 
     blockColl.put(blockHash, block) >>
       childToParentColl.put(blockHash, parentHash) >>
-      parentToChildrenColl.alter(blockHash) { maybeChildren =>
-        maybeChildren orElse Set.empty.some
-      } >>
       parentToChildrenColl.alter(parentHash) { maybeChildren =>
         maybeChildren orElse Set.empty.some map (_ + blockHash)
       }
@@ -167,13 +164,24 @@ class BlockStorage[N, A <: Agreement: Block](
         case Some((blockHash, queue)) =>
           parentToChildrenColl.read(blockHash).flatMap {
             case None =>
-              loop(queue, acc)
+              // Since we're not inserting an empty child set,
+              // we can't tell here if the block exists or not.
+              loop(queue, blockHash :: acc)
             case Some(children) =>
               loop(queue ++ children, blockHash :: acc)
           }
       }
     }
-    loop(Queue(blockHash), Nil)
+    loop(Queue(blockHash), Nil).flatMap { result =>
+      if (result.size == 1) {
+        contains(blockHash).map {
+          case true  => result
+          case false => Nil
+        }
+      } else {
+        KVStoreRead[N].pure(result)
+      }
+    }
   }
 
   /** Delete all blocks which are not descendants of a given block,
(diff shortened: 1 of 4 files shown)


metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
```diff
@@ -151,13 +160,24 @@ class BlockStorage[N, A <: Agreement: Block](
         case Some((blockHash, queue)) =>
           parentToChildrenColl.read(blockHash).flatMap {
             case None =>
-              loop(queue, acc)
+              // Since we're not inserting an empty child set,
+              // we can't tell here if the block exists or not.
+              loop(queue, blockHash :: acc)
```
rtkaczyk, line 165: If we forego adding `blockHash` to result here, then the final result could be slightly more concise:

```scala
    loop(Queue(blockHash), Nil).flatMap { result =>
      contains(blockHash).map {
        case true  => blockHash :: result
        case false => result
      }
    }
```

rtkaczyk (approved):
Looks good, just one cosmetic remark

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
aakoshh: I'll check but I think that would be different: if I don't add the `blockHash` where `parentToChildren.read(blockHash)` is `None` then I'll also not get in the result any descendant which doesn't have further children on its own, so, none of the leaves. 

Also didn't want to do another `contains` check unless it looked suspicious.

aakoshh: Unfortunately this fails many tests, or I maybe be misunderstanding what you wanted to do.

rtkaczyk: Ah, true. There's something l don't like about that check, but since it's an optimisation for `result.size == 1`, well, optimisations tend to look like that.

aakoshh: How about this?
```scala
   loop(Queue(blockHash), Nil).flatMap {
      case result @ List(`blockHash`) =>
        result.filterA(contains)
      case result =>
        KVStoreRead[N].pure(result)
    }
```

rtkaczyk: Yes, much more pleasant to the eye :+1:

PR #31, opened 2021-04-19
PM-3112: ViewState persistence

aakoshh:
Adds a `ViewStateStorage` component to store pieces of the `ProtocolState` that should survive a restart. These are the view number, the prepare/locked/commit Q.C., and the last executed block. At least I couldn't think of anything else now. (Actually there's one more thing: the root of the block tree, but I'll add that in the block synchronisation story; it will be used for pruning blocks when we can skip ahead to a new committed state without syncing blocks in between. EDIT: Added in https://github.com/input-output-hk/metronome/pull/37).

The storage is initialised by the genesis data, so it always has something to return. All items are stored under different keys, because they change at different points in time, and this way we don't have to retrieve some data, update it, then save it back, just do a single write. 

I expect that reads will happen only at startup, to initialise the `ProtocolState`, so instead of a stand-alone `ViewState` model, the reads return a `ViewStateStorage.Bundle` which is just a technical class to hold all fields together.

Also modified the `KVStore` classes to not require a `Codec` for keys since they are never decoded.

### DIFF
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
@@ -18,7 +18,10 @@ import io.iohk.metronome.hotstuff.consensus.basic.{
   QuorumCertificate
 }
 import io.iohk.metronome.hotstuff.service.pipes.BlockSyncPipe
-import io.iohk.metronome.hotstuff.service.storage.BlockStorage
+import io.iohk.metronome.hotstuff.service.storage.{
+  BlockStorage,
+  ViewStateStorage
+}
 import io.iohk.metronome.hotstuff.service.tracing.ConsensusTracers
 import io.iohk.metronome.networking.ConnectionHandler
 import io.iohk.metronome.storage.KVStoreRunner
@@ -35,6 +38,7 @@ class ConsensusService[F[_]: Timer: Concurrent, N, A <: Agreement: Block](
     publicKey: A#PKey,
     network: Network[F, A, Message[A]],
     blockStorage: BlockStorage[N, A],
+    viewStateStorage: ViewStateStorage[N, A],
     stateRef: Ref[F, ProtocolState[A]],
     stashRef: Ref[F, ConsensusService.MessageStash[A]],
     blockSyncPipe: BlockSyncPipe[F, A]#Left,
@@ -229,10 +233,39 @@ class ConsensusService[F[_]: Timer: Concurrent, N, A <: Agreement: Block](
       applySyncEffects(state, effects)
 
     // Unstash messages before we change state.
-    unstash(nextState) >>
+    captureChanges(nextState) >>
+      unstash(nextState) >>
       stateRef.set(nextState) >>
       scheduleEffects(nextEffects)
   }
+
+  /** Update the view state with and trace changes when they happen. */
+  private def captureChanges(nextState: ProtocolState[A]): F[Unit] = {
+    stateRef.get.flatMap { state =>
+      def ifChanged[T](get: ProtocolState[A] => T)(f: T => F[Unit]) = {
+        val prev = get(state)
+        val next = get(nextState)
+        f(next).whenA(prev != next)
+      }
+
+      ifChanged(_.viewNumber)(updateViewNumber) >>
+        ifChanged(_.prepareQC)(updateQuorum) >>
+        ifChanged(_.lockedQC)(updateQuorum) >>
+        ifChanged(_.commitQC)(updateQuorum)
+    }
+  }
+
+  private def updateViewNumber(viewNumber: ViewNumber): F[Unit] =
+    tracers.newView(viewNumber) >>
+      storeRunner.runReadWrite {
+        viewStateStorage.setViewNumber(viewNumber)
+      }
+
+  private def updateQuorum(quorumCertificate: QuorumCertificate[A]): F[Unit] =
+    tracers.quorum(quorumCertificate) >>
+      storeRunner.runReadWrite {
+        viewStateStorage.setQuorumCertificate(quorumCertificate)
+      }
 
   /** Requeue messages which arrived too early, but are now due becuase
     * the state caught up with them.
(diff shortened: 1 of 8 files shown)


metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/storage/ViewStateStorageProps.scala
```diff
@@ -0,0 +1,193 @@
...
+    } yield SetQuorumCertificateCommand(qc)
+
+  def genSetLastExecutedBlockHash(state: State) =
+    for {
+      h <- Gen.oneOf(
+        state.prepareQC.blockHash,
+        state.lockedQC.blockHash,
+        state.commitQC.blockHash
+      )
+    } yield SetLastExecutedBlockHashCommand(h)
```
dzajkowski, line 136:
```suggestion
  def genSetLastExecutedBlockHash(state: State) =
    Gen.oneOf(
        state.prepareQC.blockHash,
        state.lockedQC.blockHash,
        state.commitQC.blockHash
      ).map(SetLastExecutedBlockHashCommand(_))
```

dzajkowski: wouldn't this save a few lines? ;)

aakoshh: Indeed! I think I just wanted to match the visual style of the other two generators above it which `yield` a command.

aakoshh: Annoyingly `scalafmt` decided that we cannot go from 7 to 5 lines :disappointed: 
```scala
    Gen
      .oneOf(
        state.prepareQC.blockHash,
        state.lockedQC.blockHash,
        state.commitQC.blockHash
      )
      .map(SetLastExecutedBlockHashCommand(_))
```

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
```diff
@@ -229,11 +233,40 @@ class ConsensusService[F[_]: Timer: Concurrent, N, A <: Agreement: Block](
       applySyncEffects(state, effects)
 
     // Unstash messages before we change state.
-    unstash(nextState) >>
+    captureChanges(nextState) >>
+      unstash(nextState) >>
```
rtkaczyk, line 237: I'm a bit confused by the comment above. Does the order really matter here?

rtkaczyk: Ahh, because unstash operates on the current `stateRef` :+1:

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/ViewStateStorage.scala
```diff
@@ -0,0 +1,148 @@
...
+    object Key {
+      case object ViewNumber            extends Key[ViewNumber]
+      case object PrepareQC             extends Key[QuorumCertificate[A]]
+      case object LockedQC              extends Key[QuorumCertificate[A]]
+      case object CommitQC              extends Key[QuorumCertificate[A]]
+      case object LastExecutedBlockHash extends Key[A#Hash]
+
+      implicit def encoder[V]: Encoder[Key[V]] =
+        scodec.codecs.implicits.implicitStringCodec
+          .contramap[Key[V]](_.toString)
```
rtkaczyk, line 74: Aren't we concerned with storage efficiency? How about using something like `sealed abstract class Key[A <: Agreement](val code: Int)` and using that `code` as storage key?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
```diff
@@ -357,8 +387,7 @@ class ConsensusService[F[_]: Timer: Concurrent, N, A <: Agreement: Block](
         // should be offloaded to another queue.
         //
         // Save the Commit Quorum Certificate to the view state.
-        saveCommitQC(commitQC) >>
-          blockExecutionQueue.offer(effect)
+        blockExecutionQueue.offer(effect)
```
rtkaczyk, line 390: The comment about saving the QC and the pattern match are now obsolete, right?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
aakoshh: Yeah, I don't know if I should have threaded the state itself to all these methods, seemed more awkward.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/ViewStateStorage.scala
aakoshh: I don't know, there's only a couple of keys. Can't it handle some strings?

I'm concerned with one of @ntallar comments in the other project that reinserting the same records over and over caused RocksDB to grow until it was restarted, like, space after deleted records were not reclaimed automatically.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
aakoshh: True, thanks!

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/ViewStateStorage.scala
aakoshh: There must be some sort of concern to cause people to use [single letter namespaces](https://github.com/input-output-hk/mantis/blob/develop/src/main/scala/io/iohk/ethereum/db/storage/Namespaces.scala#L4-L14), which got really ugly in the other project where we ran out of intuitive letters and started going down the alphabet. Numbers would be much better in that sense, but I thought this even more foolproof and lazy.

aakoshh: But you may be right, why not. Pushed a change to use 1 byte.

rtkaczyk (approved):

PR #32, opened 2021-04-20
PM-3134: Block synchronisation

aakoshh:
Required PRs are https://github.com/input-output-hk/metronome/pull/29, https://github.com/input-output-hk/metronome/pull/28, https://github.com/input-output-hk/metronome/pull/25 and https://github.com/input-output-hk/metronome/pull/31

* Added an `RPCTracker` component that can match (optional) responses to requests, or issue a timeout.
* Added `BlockSynchronizer` to download and persist blocks that have a Quorum Certificate.
* Wired together `SyncService` and `BlockSynchronizer`. 
* Enhanced `FiberMap` with the ability the cancel the enqueued tasks of a key.

The idea is that we have a block tree in persistent storage, but sometimes a `Prepare` message can contain a block we don't have a parent for. But the `Prepare` message contains the Quorum Certificate for that parent, so we know that it's already been vetted by the federation and it's safe to download and store - we don't even need to validate it, similarly to how we can just adopt any state that has a Commit Q.C. for it without validating its history. 

The `BlockSynchronizer` downloads ancestors of the High Q.C. until it finds a block that is already in persistent storage. The blocks it's downloading go in an in-memory block store until they can be connected to the persistent tree, at which point they are moved over. This way we never risk ending up with a forest in the persistent store if the program crashes in the middle of a download sequence, which would make pruning difficult later. 

However, this means we are limited in how much we can download by the amount of memory we have. We made it an assumption that this block-by-block download is happening only when we're in-sync with the other federation members and the number of missing blocks is likely to be very small. Otherwise the HotStuff service will recognise that it's out of sync and trigger state synchronisation, which polls the members for the highest commit Q.C. and requests a jump directly there. So the overall application is expected to be able to do _something_ to carry out a faster than normal synchronisation to any given block hash, which will not use the mechanism in the `BlockSynchronizer`. 

When it's time to switch from regular block synchronisation to jump synchronisation, the `SyncService` will be equipped with the means to cancel all outstanding downloads, sync the state, prune the existing block store, re-kindle it from the new block, then resume normal sync.

The `BlockSynchronizer` itself is simplistic at the moment: in each round we can start downloading one more High Q.C. from whoever the current leader is. We could expect other members of the federation to have this block as well, (although we don't necessarily know who signed the Q.C. if we use threshold cryptography), however, currently adding an alternative source comes naturally if the next leader also builds on the same, or a descendant High Q.C. At that point multiple fibers will be ascending the same chain, potentially downloading any missing block from multiple sources, until one of them succeeds. Only one request is issued to any given federation member at a time.

### DIFF
--- a/metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
+++ b/metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
@@ -1,0 +1,89 @@
+package io.iohk.metronome.core.messages
+
+import cats.implicits._
+import cats.effect.{Concurrent, Timer, Sync}
+import cats.effect.concurrent.{Ref, Deferred}
+import java.util.UUID
+import scala.concurrent.duration.FiniteDuration
+import scala.reflect.ClassTag
+
+/** `RPCTracker` can be used to register outgoing requests and later
+  * match them up with incoming responses, thus turning the two independent
+  * messages into a `Kleisli[F, Request, Option[Response]]`, where a `None`
+  * result means the request timed out.
+  */
+class RPCTracker[F[_]: Timer: Concurrent, M](
+    deferredMapRef: Ref[F, Map[UUID, RPCTracker.Entry[F, _]]],
+    defaultTimeout: FiniteDuration
+) {
+  import RPCTracker.Entry
+
+  def register[
+      Req <: RPCMessageCompanion#Request,
+      Res <: RPCMessageCompanion#Response
+  ](
+      request: Req,
+      timeout: FiniteDuration = defaultTimeout
+  )(implicit
+      ev1: Req <:< M,
+      ev2: RPCPair.Aux[Req, Res],
+      ct: ClassTag[Res]
+  ): F[F[Option[Res]]] = {
+    val requestId = request.requestId
+    for {
+      d <- Deferred[F, Option[Res]]
+      e = RPCTracker.Entry(d)
+      _ <- deferredMapRef.update(_ + (requestId -> e))
+      _ <- Concurrent[F].start {
+        Timer[F].sleep(timeout) >> completeWithTimeout(requestId)
+      }
+    } yield d.get
+  }
+
+  def complete[Res <: RPCMessageCompanion#Response](
+      response: Res
+  )(implicit ev: Res <:< M): F[Boolean] = {
+    remove(response.requestId).flatMap {
+      case None    => false.pure[F]
+      case Some(e) => e.complete(response)
+    }
+  }
+
+  private def completeWithTimeout(requestId: UUID): F[Unit] =
+    remove(requestId).flatMap {
+      case None    => ().pure[F]
+      case Some(e) => e.timeout
+    }
+
+  private def remove(requestId: UUID): F[Option[Entry[F, _]]] =
+    deferredMapRef.modify { dm =>
+      (dm - requestId, dm.get(requestId))
+    }
+}
+
+object RPCTracker {
+  case class Entry[F[_]: Sync, Res](
+      deferred: Deferred[F, Option[Res]]
+  )(implicit ct: ClassTag[Res]) {
+    def timeout: F[Unit] =
+      deferred.complete(None).attempt.void
+
+    def complete[M](response: M): F[Boolean] = {
+      response match {
+        case expected: Res =>
+          deferred.complete(Some(expected)).attempt.map(_.isRight)
+        case _ =>
+          // Not returning an error because if the message arrived after
+          // the timeout we wouldn't know anyway what it was meant to complete.
+          false.pure[F]
+      }
+    }
+  }
+
+  def apply[F[_]: Concurrent: Timer, M](
+      defaultTimeout: FiniteDuration
+  ): F[RPCTracker[F, M]] =
+    Ref[F].of(Map.empty[UUID, Entry[F, _]]).map {
+      new RPCTracker(_, defaultTimeout)
+    }
+}
--- a/metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
+++ b/metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
@@ -52,11 +52,18 @@ class RPCTracker[F[_]: Timer: Concurrent, M](
     } yield d.get
   }
 
+  /** Try to complete an outstanding request with a response.
+    *
+    * Returns `true` if the response was expected, `false` if
+    * it wasn't, or already timed out. An error is returned
+    * if the response was expected but the there was a type
+    * mismatch.
+    */
   def complete[Res <: RPCMessageCompanion#Response](
       response: Res
-  )(implicit ev: Res <:< M): F[Boolean] = {
+  )(implicit ev: Res <:< M): F[Either[Throwable, Boolean]] = {
     remove(response.requestId).flatMap {
-      case None    => false.pure[F]
+      case None    => false.asRight[Throwable].pure[F]
       case Some(e) => e.complete(response)
     }
   }
@@ -80,15 +87,23 @@ object RPCTracker {
     def timeout: F[Unit] =
       deferred.complete(None).attempt.void
 
-    def complete[M](response: M): F[Boolean] = {
+    def complete[M](response: M): F[Either[Throwable, Boolean]] = {
       response match {
         case expected: Res =>
-          deferred.complete(Some(expected)).attempt.map(_.isRight)
+          deferred
+            .complete(Some(expected))
+            .attempt
+            .map(_.isRight.asRight[Throwable])
         case _ =>
           // Wrong type, as evidenced by `ct` not maching `Res`.
-          // Not returning an error because if the message arrived after
-          // the timeout we wouldn't know anyway what it was meant to complete.
-          deferred.complete(None).attempt.as(false)
+          // Returning an error so that this kind of programming error
+          // can be highlighted as soon as possible. Note though that
+          // if the request already timed out we can't tell if this
+          // error would have happened if the response arrived earlier.
+          val error = new IllegalArgumentException(
+            s"Invalid response type ${response.getClass.getName}; expected ${ct.runtimeClass.getName}"
+          )
+          deferred.complete(None).attempt.as(error.asLeft[Boolean])
       }
     }
   }
--- a/metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
+++ b/metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
@@ -8,9 +8,19 @@ import scala.concurrent.duration.FiniteDuration
 import scala.reflect.ClassTag
 
 /** `RPCTracker` can be used to register outgoing requests and later
-  * match them up with incoming responses, thus turning the two independent
-  * messages into a `Kleisli[F, Request, Option[Response]]`, where a `None`
-  * result means the request timed out.
+  * match them up with incoming responses, thus it facilitates turning
+  * the two independent messages into a `Kleisli[F, Request, Option[Response]]`,
+  * by a component that has access to the network, where a `None` result means
+  * the operation timed out before a response was received.
+  *
+  * The workflow is:
+  * 0. Receive some request parameters in a method.
+  * 1. Create a request ID.
+  * 2. Create a request with the ID.
+  * 3. Register the request with the tracker, hold on to the handle.
+  * 4. Send the request over the network.
+  * 5. Wait on the handle, eventually returning the optional result to the caller.
+  * 6. Pass every response received from the network to the tracker (on the network handler fiber).
   */
 class RPCTracker[F[_]: Timer: Concurrent, M](
     deferredMapRef: Ref[F, Map[UUID, RPCTracker.Entry[F, _]]],
@@ -27,6 +37,8 @@ class RPCTracker[F[_]: Timer: Concurrent, M](
   )(implicit
       ev1: Req <:< M,
       ev2: RPCPair.Aux[Req, Res],
+      // Used by `RPCTracker.Entry.complete` to make sure only the
+      // expected response type can complete a request.
       ct: ClassTag[Res]
   ): F[F[Option[Res]]] = {
     val requestId = request.requestId
@@ -73,9 +85,10 @@ object RPCTracker {
         case expected: Res =>
           deferred.complete(Some(expected)).attempt.map(_.isRight)
         case _ =>
+          // Wrong type, as evidenced by `ct` not maching `Res`.
           // Not returning an error because if the message arrived after
           // the timeout we wouldn't know anyway what it was meant to complete.
-          false.pure[F]
+          deferred.complete(None).attempt.as(false)
       }
     }
   }
--- a/metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
+++ b/metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
@@ -5,10 +5,11 @@ import monix.execution.Scheduler.Implicits.global
 import org.scalatest.flatspec.AsyncFlatSpec
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.compatible.Assertion
+import org.scalatest.Inside
 import scala.concurrent.Future
 import scala.concurrent.duration._
 
-class RPCTrackerSpec extends AsyncFlatSpec with Matchers {
+class RPCTrackerSpec extends AsyncFlatSpec with Matchers with Inside {
 
   sealed trait TestMessage extends RPCMessage
   object TestMessage extends RPCMessageCompanion {
@@ -44,7 +45,7 @@ class RPCTrackerSpec extends AsyncFlatSpec with Matchers {
       ok   <- tracker.complete(res)
       got  <- join
     } yield {
-      ok shouldBe true
+      ok shouldBe Right(true)
       got shouldBe Some(res)
     }
   }
@@ -52,11 +53,14 @@ class RPCTrackerSpec extends AsyncFlatSpec with Matchers {
   it should "complete responses with None after the timeout" in test {
     tracker =>
       val req = FooRequest(RequestId())
+      val res = FooResponse(req.requestId, 1)
       for {
         join <- tracker.register(req, timeout = 50.millis)
         _    <- Task.sleep(100.millis)
+        ok   <- tracker.complete(res)
         got  <- join
       } yield {
+        ok shouldBe Right(false)
         got shouldBe empty
       }
   }
@@ -71,7 +75,9 @@ class RPCTrackerSpec extends AsyncFlatSpec with Matchers {
         ok   <- tracker.complete(res)
         got  <- join
       } yield {
-        ok shouldBe false
+        inside(ok) { case Left(error) =>
+          error.getMessage should include("Invalid response type")
+        }
         got shouldBe empty
       }
   }
--- a/metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
+++ b/metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
@@ -1,0 +1,78 @@
+package io.iohk.metronome.core.messages
+
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.compatible.Assertion
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+class RPCTrackerSpec extends AsyncFlatSpec with Matchers {
+
+  sealed trait TestMessage extends RPCMessage
+  object TestMessage extends RPCMessageCompanion {
+    case class FooRequest(requestId: RequestId) extends TestMessage with Request
+    case class FooResponse(requestId: RequestId, value: Int)
+        extends TestMessage
+        with Response
+    case class BarRequest(requestId: RequestId) extends TestMessage with Request
+    case class BarResponse(requestId: RequestId, value: String)
+        extends TestMessage
+        with Response
+
+    implicit val foo = pair[FooRequest, FooResponse]
+    implicit val bar = pair[BarRequest, BarResponse]
+  }
+  import TestMessage._
+
+  def test(
+      f: RPCTracker[Task, TestMessage] => Task[Assertion]
+  ): Future[Assertion] =
+    RPCTracker[Task, TestMessage](10.seconds)
+      .flatMap(f)
+      .timeout(5.seconds)
+      .runToFuture
+
+  behavior of "RPCTracker"
+
+  it should "complete responses within the timeout" in test { tracker =>
+    val req = FooRequest(RequestId())
+    val res = FooResponse(req.requestId, 1)
+    for {
+      join <- tracker.register(req)
+      ok   <- tracker.complete(res)
+      got  <- join
+    } yield {
+      ok shouldBe true
+      got shouldBe Some(res)
+    }
+  }
+
+  it should "complete responses with None after the timeout" in test {
+    tracker =>
+      val req = FooRequest(RequestId())
+      for {
+        join <- tracker.register(req, timeout = 50.millis)
+        _    <- Task.sleep(100.millis)
+        got  <- join
+      } yield {
+        got shouldBe empty
+      }
+  }
+
+  it should "complete responses with None if the wrong type of response arrives" in test {
+    tracker =>
+      val req = FooRequest(RequestId())
+      val res = BarResponse(RequestId(), "one")
+      for {
+        join <- tracker.register(req, timeout = 50.millis)
+        ok   <- tracker.complete(res)
+        got  <- join
+      } yield {
+        ok shouldBe false
+        got shouldBe empty
+      }
+  }
+
+}
--- a/metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
+++ b/metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
@@ -63,10 +63,11 @@ class RPCTrackerSpec extends AsyncFlatSpec with Matchers {
 
   it should "complete responses with None if the wrong type of response arrives" in test {
     tracker =>
-      val req = FooRequest(RequestId())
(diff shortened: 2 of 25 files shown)


metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  def register[
+      Req <: RPCMessageCompanion#Request,
+      Res <: RPCMessageCompanion#Response
+  ](
+      request: Req,
+      timeout: FiniteDuration = defaultTimeout
+  )(implicit
+      ev1: Req <:< M,
+      ev2: RPCPair.Aux[Req, Res],
```
rtkaczyk, line 29: I'm a bit confused:
- how is this evidence used?
- why do we need `Aux`? Wouldn't `type RPCPair[Req, Res]` suffice?
- is there going to be one tracker for all sync messages, or one for each concrete pair like `GetBlock*`?

metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+        got  <- join
+      } yield {
+        got shouldBe empty
+      }
+  }
+
+  it should "complete responses with None if the wrong type of response arrives" in test {
+    tracker =>
+      val req = FooRequest(RequestId())
+      val res = BarResponse(RequestId(), "one")
```
rtkaczyk, line 67: If I understand correctly the failed `complete` is not really due to the wrong type but due to a different UUID. In fact, if the Ids matched the request would be completed with the wrong response type, would it not?

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
```diff
@@ -0,0 +1,89 @@
...
+import cats.effect.concurrent.{Ref, Deferred}
+import java.util.UUID
+import scala.concurrent.duration.FiniteDuration
+import scala.reflect.ClassTag
+
+/** `RPCTracker` can be used to register outgoing requests and later
+  * match them up with incoming responses, thus turning the two independent
+  * messages into a `Kleisli[F, Request, Option[Response]]`, where a `None`
+  * result means the request timed out.
+  */
```
rtkaczyk, line 14: I understand you mean that this is _equivalent_ to a Kleisli, but I still fail to see the similarity. Could you explain please?

aakoshh: Sorry, that was a bit vague, it doesn't turn them into a `Request => F[Option[Response]]` per se, but it facilitates that pattern by a component that does network communication.

What I meant is that we have this `Network` interface where you have a stream of incoming messages, and you can send messages in a fashion similar to an Akka Actor, i.e. `send: Message => F[Unit]`. But in many places it would be useful if we could send a request and wait for the response. We can do that by adding a listener to the incoming stream and matching the `requestId`.

This `RPCTracker` allows us to do that: every `Request` has a `requestId`, which a `Response` has to echo back, and the tracker will register an outgoing request (has to be done before it's sent away), and later can handle the `Response` by completing the equivalent of a `Deferred`. It also handles timeouts, so we never wait forever. The `F[F[Option[Response]]]` returned by `submit` can be used to wait for the response, or the timeout.

So`RPCTracker` is used by the `SyncService` to present network communication as a Kleisli.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
```diff
@@ -0,0 +1,210 @@
...
+      blockHash: A#Hash,
+      path: List[A#Hash]
+  ): F[List[A#Hash]] = {
+    storeRunner
+      .runReadOnly {
+        blockStorage.contains(blockHash)
+      }
+      .flatMap {
+        case true =>
+          path.pure[F]
```
rtkaczyk, line 83: Are we guaranteed that if the block exists in storage then so do its ancestors?

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
aakoshh: > how is this evidence used?

It's here so the compiler can check that if we send a given `Req` type then we are right to expect a `Res` to be returned. Not just that, but should also allow the compiler to infer the return type based on just the `request` parameter, without explicitly having to define both generic types for `register`. 

> why do we need Aux? Wouldn't type RPCPair[Req, Res] suffice?

Every request has one possible response (although I suppose the same response could be paired with multiple responses), so `RPCPair` is just defined in terms of `Req`, with `Res` being a dependant type. Having two parameters would allow the same request to have multiple responses.

The `Aux` is needed so we can match against the two generic types of the `register` method. 

Otherwise it could be like this:
```scala
def register[
      Req <: RPCMessageCompanion#Request
  ](
      request: Req,
      timeout: FiniteDuration = defaultTimeout
  )(implicit
      ev1: Req <:< M,
      ev2: RPCPair[Req]
  ): F[F[Option[ev2.Res]]]
```
But then I couldn't have a `ct: ClassTag[Res]` because it cannot be matched against a dependent type of a parameter, `ev2`.

> is there going to be one tracker for all sync messages, or one for each concrete pair like GetBlock*?

Just one tracker per "service", like the `SyncService`.

metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
aakoshh: Good point about the request ID, but it wouldn't be completed with the wrong type because of the `ClassTag` which is passed to `register`, it's used to only match responses of the expected type.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
aakoshh: They could be pruned, but we'll take care that we have the last executed and committed block, so we can extend it.

metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
aakoshh: Fixed the test. It turns out the code didn't do what the test expected, it allowed the timeout to happen rather than complete the operation with `None`. I thought which one's the better behaviour, but fixed it so `None` is returned immediately because:
* A malicious user can try to stall us by just not sending anything, they have no incentive to send the wrong kind of data.
* If we get the wrong kind of data it's most likely down to some programming error, very unlikely that a correct response will follow.
* Maybe it's an error with a certain node, and returning immediately gives us the option to move on and ask somebody else.

But I may be wrong, maybe allowing it to time out and a potentially correct message to arrive is better.

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
aakoshh: Added comments.

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
aakoshh: Changed the wording and added a workflow description.

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
rtkaczyk: Thanks for the explanation

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
```diff
@@ -0,0 +1,102 @@
...
+      deferred.complete(None).attempt.void
+
+    def complete[M](response: M): F[Boolean] = {
+      response match {
+        case expected: Res =>
+          deferred.complete(Some(expected)).attempt.map(_.isRight)
+        case _ =>
+          // Wrong type, as evidenced by `ct` not maching `Res`.
+          // Not returning an error because if the message arrived after
+          // the timeout we wouldn't know anyway what it was meant to complete.
```
rtkaczyk, line 90: Perhaps the error wouldn't be that useful, but I don't quite get the argumentation. This code path isn't run when a timeout occurs so we would be able to differentiate those cases.

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
```diff
@@ -0,0 +1,102 @@
...
+      (dm - requestId, dm.get(requestId))
+    }
+}
+
+object RPCTracker {
+  case class Entry[F[_]: Sync, Res](
+      deferred: Deferred[F, Option[Res]]
+  )(implicit ct: ClassTag[Res]) {
+    def timeout: F[Unit] =
+      deferred.complete(None).attempt.void
```
rtkaczyk, line 81: Perhaps a timeout is worth a tracer? Will we be able to add it?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
```diff
@@ -27,28 +39,57 @@ class SyncService[F[_]: Sync, N, A <: Agreement](
...
-    * Returns `None` if we're not connected or the request times out.
-    */
-  def getBlock(from: A#PKey, blockHash: A#Hash): F[Option[A#Block]] = ???
+  /** Request a block from a peer. */
+  private def getBlock(from: A#PKey, blockHash: A#Hash): F[Option[A#Block]] = {
+    for {
+      requestId <- RequestId[F]
+      request = GetBlockRequest(requestId, blockHash)
+      maybeResponse <- sendRequest(from, request)
+    } yield maybeResponse.map(_.block)
```
rtkaczyk, line 54: Could you explain why using a `for` makes this referentially transparent in contrast to the previous version? (just curious)

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
aakoshh: Right, that wasn't very clear. What I meant was that if a timeout occurred independently and earlier than `tracker.complete(wrongResponse)` is called, then the `Entry` would be gone by this time and we couldn't differentiate why the response hasn't been handled, it would look like any other timeout. 

But you may be right, it could be worth differentiating anyway, just to catch these problems earlier.

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
aakoshh: I don't want to add it to the `RPCTracker` itself, because in `core` there's no dependency on `tracing`, and currently tracers are on a bit coarser granularity (service level), but there's indeed [a trace event](https://github.com/input-output-hk/metronome/blob/1447f0bb0990d751223fe1e24d9c8691658a3fc4/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala#L102) for timeouts in the `SyncService`.

aakoshh: There's also one for [ignored responses](https://github.com/input-output-hk/metronome/blob/1447f0bb0990d751223fe1e24d9c8691658a3fc4/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala#L159) so you can investigate why a certain response type always times out and always ignored, _and_ at the same time -> conclude that it's the wrong type.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Sure, it was the change from `RequestId()` to `RequestId[F]`

Previously it was something like this:
```scala
def send(params): F[Response] = {
  val requestId = RequestId()
  val request = Request(requestId, params)
  Sync[F].delay(rpc(request))
}
```
So when we call `send` we always get a different request ID, so these two were different:
```scala
val a = send(1)
val b = a // Uses the same request ID
// vs
val a = send(1)
val b = send(1) // Uses a different request ID.
```

Now it's like this:
```scala
def send(params): F[Response] = 
  RequestId[F].map(Request(_, params)).flatMap(rpc)
```
When we call `send`, no side effect happens, so the two cases above would be equivalent.

metronome/core/src/io/iohk/metronome/core/messages/RPCTracker.scala
aakoshh: Added different handling for errors due to type mismatch.

metronome/core/test/src/io/iohk/metronome/core/messages/RPCTrackerSpec.scala
rtkaczyk: I also prefer returning `None`. It should be easier to catch any errors and I don't think it can be abused

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
```diff
@@ -98,43 +137,63 @@ class SyncService[F[_]: Sync, N, A <: Agreement](
...
+          // If the same leader is sending us newer proposals, we can ignore the
+          // previous pepared blocks - they are either part of the new Q.C.,
+          // in which case they don't need to be validated, or they have not
+          // gathered enough votes, and been superseded by a new proposal.
+          syncFiberMap.cancelQueue(sender) >>
+            syncFiberMap
+              .submit(sender) {
+                for {
+                  _       <- blockSynchronizer.sync(sender, prepare.highQC)
+                  isValid <- validateBlock(prepare.block)
```
rtkaczyk, line 184: Does this realise the `safeNode` predicate? If so, it probably should depend on `highQC` as well. Just asking, it could be added in the relevant tasks

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
```diff
@@ -0,0 +1,210 @@
...
+        Block[A].blockHash(block) == blockHash &&
+        Block[A].isValid(block)
+      }
+    }
+
+  /** See how far we can go in memory from the original block hash we asked for,
+    * which indicates the blocks that no concurrent download has persisted yet,
+    * then persist the rest.
+    *
+    * Only doing oinepersist operation at a time to make sure there's no competition
```
rtkaczyk, line 136: typo: one persist

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
```diff
@@ -0,0 +1,210 @@
...
+      blockHash: A#Hash,
+      path: List[A#Hash]
+  ): F[Unit] =
+    semaphore.withPermit {
+      inMemoryStore
+        .runReadOnly {
+          blockStorage.getPathFromRoot(blockHash).map(_.toSet)
+        }
+        .flatMap { unpersisted =>
+          persist(path, unpersisted)
```
rtkaczyk, line 149: Since this is guarded by a semaphore I think it should hold that `unpersisted.isEmpty || path.toSet == unpersisted`. So how about:

```
// no .map(_.toSet)
.flatMap {
  case Nil => ().pure[F]
  case unpersisted => persistAndClear(unpersisted)
}
```

I would use a different name for the inner function. At least my brain is easily tricked towards seeing a recursion.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: No, that's not here, this is only validating whether the block contents are valid according to the application rules, so this is where we'd ask the PoW Interpreter whether any checkpoint candidates in the block are valid in the context of the state of the parent block (which has been validated before, evidenced by the High Q.C.).

The `safeNode` predicate is checked in [ProtocolState](https://github.com/input-output-hk/metronome/blob/develop/metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/basic/ProtocolState.scala#L194) during the `Prepare` phase. That happens in `handleMessage` which already receives a validated block. 

There is some preliminary stateless validation in `ProtocolState.validateMessage`, but because that's stateless it doesn't check the extension rule (`lockedQC` depends on consensus state).

However, I added two more extension validations to the `ConsensusService`:
1. There's a [view number check](https://github.com/input-output-hk/metronome/blob/aa9e54de0a72a5db80650db1b88628abd78dd9fd/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala#L88-L97) when the message is received to see that the leader is not proposing to use a quorum certificate that is older than our committed one.
2. There's a [parent check](https://github.com/input-output-hk/metronome/blob/aa9e54de0a72a5db80650db1b88628abd78dd9fd/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala#L128-L131) before we hand over to the `SyncService` to see that the proposed block actually extends the High Q.C.

Based on these we can tell whether the block proposed to us looks valid extension of its parent and the previously committed state. Then the `SyncService` gets any missing dependencies and validates the content, replying with the outcome. Only then is the block delivered to the `ProtocolState` for consensus decision. By that time we may have moved on to a new view, but if everything checks out and it's a safe extension, a vote will be cast to it. 

This passing around the message realises this passage from the paper:

> During the protocol, a replica delivers a message only after the branch led by the node is already in its local tree.  In practice, a recipient who falls behind can catch up by fetching missing nodes from other replicas. 

In other words, by the time the consensus model sees the block, they should have already been vetted for their content. I'm just doing those extra 2 checks in the `ConsensusService` to throw out spam.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
aakoshh: Sure, happy to rename, I had to look hard what the difference was now as well. I wouldn't move the `Nil` here though, it's the first thing the other method does anyway, it's okay to call it with an empty list, since that's what it does anyway. Prefer to have check of the same condition. 

`unpersisted` and `path` are not necessarily the same. Here's a scenario:

```
                   B5 - B7
                  /
B0* - B1* - B2 - B4 - B6
             \ 
              B3
```
Say that B1 is our last persisted (marked by `*`) block, and we have 3 proposal coming in: B3, B7 and B6, all enqueued for download at the same time. This should result in 3 paths: `[B2, B3]; [B2, B4, B5, B7]; [B2, B4, B6]`. These blocks will all be added to the in memory block store as we walk backwards toward the persistent part, ending at B1. 

Then we call `persist`, which will remove the blocks from the in-memory store and move them to persistent storage in the order they appear in the `path`. We only do one at a time because of the semaphore, but the order can be arbitrary. Say we start with the last path.

After persisting `[B2, B4, B6]` the in-memory store will not have any of them any more. Take the 2nd path next. We start by determining which blocks are still unpersisted by traversing from B7 back on the parent path in the in-memory tree; this gives us an `unpersisted` path of `[B5, B7]`, which is not the same as the original we had to traverse back to B2.

I can't remember why I wasn't using just `unpersisted` as a path though, why use it as a filter while traversing `path`. Maybe I didn't have the semaphore at the time. :thinking:

aakoshh: Oh yeah, I know what I meant by this:
> The path is maintained separately from the in-memory store in case another ongoing download would re-insert something on a path already partially removed resulting in a forest that cannot be traversed fully.

While `persist` is protected by a semaphore, the concurrent downloads are free to _insert_ into the in-memory tree. I was thinking about a scenario where download 1 is persisting and removing block while download 2 is re-inserting them during traversal. 

Take the previous scenario. Say we started downloading B6 and B7 at the same time, from different sources, Alice and Bob, and due to timing what happens is that we start downloading B2 from Alice and Bob at the same time. Say Alice's finishes first and we start persisting the path `[B2, B4, B6]`: we remove B2 from the in-memory tree, persist it, then B4, then B6. During this time, the download of B2 from Bob finishes and is re-inserted into the in-memory tree.

When Bob considers what he can persist, the in memory tree has B2, B5 and B7 in it, so it's not a tree any more, but a forest, B4 is missing. If Bob only checked how far he has can traverse from B7 up the tree, he'd only persist B5 and B7, and not remove B2 from the in-memory store, leaking memory. 

That's why Bob will always try to persist `[B2, B4, B5, B7]` but be advised that only B5 and B7 actually need insertion into the persistent storage, B2 must already be there because it's unreachable, so B2 is just removed from the in-memory store, and B4 does nothing because it's not in there.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: This method has been [farmed out](https://github.com/input-output-hk/metronome/blob/aa9e54de0a72a5db80650db1b88628abd78dd9fd/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala#L244) to the `ApplicationService` since this PR, so it's a bit clearer what the scope is.

rtkaczyk: Thanks for clarifying

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
rtkaczyk: Nice explanation, thank you

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
```diff
@@ -0,0 +1,218 @@
...
+                getAndValidateBlock(from, blockHash)
+                  .flatMap {
+                    case Some(block) =>
+                      inMemoryStore.runReadWrite {
+                        blockStorage.put(block)
+                      } >> downloadParent(from, block, path)
+
+                    case None =>
+                      Timer[F].sleep(retryTimeout) >>
+                        download(from, blockHash, path)
```
rtkaczyk, line 104: Currently `sync` is only called in response to `prepare` message and we are trying to sync from only the sender (leader). If the sender becomes unresponsive this will loop forever. If we sync from the replicas as well (as you mention in another comment), then it should resolve this (someone should ultimately respond or it's us who are the problem). So when are we going to sync from the replicas?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizerProps.scala
```diff
@@ -0,0 +1,196 @@
...
+    // Simulate a some random time, which may or may not be enough to finish the downloads.
+    scheduler.tick(duration)
+
+    // Check now that there the persistent store has just one tree.
+    val test = for {
+      persistent <- fixture.persistentRef.get
+    } yield {
+      persistent(Namespace.Blocks).forall { case (_, block: TestBlock) =>
+        block.parentId.isEmpty ||
+          persistent(Namespace.Blocks).contains(block.parentId)
```
rtkaczyk, line 185: This could use comment because it's not obvious that this is a valid condition for a tree, unless one looks at the generator

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
aakoshh: Yeah, I didn't want to complicate it with a download manager that has a queue and tracks the available sources, so at the moment the way it works is that we only ever try to download blocks that have a Quorum Certificate, so they are not completely rubbish, nobody can take us down the garden path with a block that has parents that don't lead anywhere. But I quite possibly missed something.

Say Alice is the leader this round and she sends us block B7 to prepare, but for whatever reason we are missing B5, so we start downloading that from Alice. My assumption was that if for whatever reason this keeps failing, the next leader, Bob, will propose something else built on B5, and we'll just get it from him.

But if he's using a different prepare quorum certificate then indeed B5 may never again be build upon and we might never get it from Alice. 

I guess it would be prudent to add some clause to give up after a while and just respond that the block proposed by Alice is not valid, or never even respond to that in the `SyncPipe`.

aakoshh: Hm, unfortunately the `BlockSynchronizer` is a bit simplistic in this regard: because we have already inserted whatever we _could_ download from Alice into the in-memory tree, and we don't know what else might depend on them, we can't just remove them on our way out, because it could clash with the `unpersisted` mechanism described above. 

Our best option is to ask another federation member.

rtkaczyk: Should we add a TODO comment to reconsider it later?

aakoshh: Pushed a change that asks the others in a randomized order. That should not fail, but if it does, it is going to retry forever until somebody serves the block.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizerProps.scala
aakoshh: Added :+1:

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
rtkaczyk: Interesting, I like it :+1:

rtkaczyk (approved):
Thanks for the patience in answering my questions :)

aakoshh:
And thank _you_ for keeping me honest ;)

PR #34, opened 2021-04-23
PM-3105: Ledger persistence

aakoshh:
Adds a `LedgerStorage` component that stores a certain number of the latest ledger instances in a ring buffer. The idea is that we only need to keep the last committed state, however we want to leave some time for other nodes that may need to sync with us to request it, so we can configure a lag after which the oldest ledger is overwritten. 

It is located in `checkpointing.service.storages` so it doesn't have to deal with anything generic, we know the ledger is going to be small. 

Builds on https://github.com/input-output-hk/metronome/pull/31 so keys don't need `Codec` just `Encoder`.

### DIFF
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
@@ -1,9 +1,8 @@
 package io.iohk.metronome.checkpointing.service.storage
 
-import cats.implicits._
 import io.iohk.metronome.checkpointing.models.Ledger
-import io.iohk.metronome.storage.{KVCollection, KVStore, KVStoreRead}
-import scodec.{Decoder, Encoder, Codec}
+import io.iohk.metronome.storage.{KVRingBuffer, KVCollection, KVStore}
+import scodec.Codec
 
 /** Storing the committed and executed checkpoint ledger.
   *
@@ -26,23 +25,12 @@ class LedgerStorage[N](
     ledgerColl: KVCollection[N, Ledger.Hash, Ledger],
     ledgerMetaNamespace: N,
     maxHistorySize: Int
-)(implicit codecH: Codec[Ledger.Hash]) {
-  require(maxHistorySize > 0, "Has to store at least one ledger.")
-
-  import LedgerStorage._
-  import scodec.codecs.implicits.implicitIntCodec
-
-  private implicit val kvn = KVStore.instance[N]
-
-  private def getMetaData[V: Decoder](key: MetaKey[V]) =
-    KVStore[N].get[MetaKey[V], V](ledgerMetaNamespace, key)
-
-  private def putMetaData[V: Encoder](key: MetaKey[V], value: V) =
-    KVStore[N].put(ledgerMetaNamespace, key, value)
-
-  /** Return the index of the next bucket to write the data into. */
-  private def nextIndex(maybeIndex: Option[Int]): Int =
-    maybeIndex.fold(0)(index => (index + 1) % maxHistorySize)
+)(implicit codecH: Codec[Ledger.Hash])
+    extends KVRingBuffer[N, Ledger.Hash, Ledger](
+      ledgerColl,
+      ledgerMetaNamespace,
+      maxHistorySize
+    ) {
 
   /** Save a new ledger and remove the oldest one, if we reached
     * the maximum history size. Since we only store committed
@@ -50,46 +38,5 @@ class LedgerStorage[N](
     * by going through a block pointing at them directly.
     */
   def put(ledger: Ledger): KVStore[N, Unit] =
-    for {
-      index <- getMetaData(BucketIndex).map(nextIndex)
-      ledgerHash = ledger.hash
-      maybeOldestHash <- getMetaData(Bucket(index))
-      _ <- maybeOldestHash match {
-        case Some(oldestHash) if oldestHash == ledgerHash =>
-          KVStore[N].unit
-
-        case Some(oldestHash) =>
-          ledgerColl.put(ledgerHash, ledger) >>
-            ledgerColl.delete(oldestHash)
-
-        case None =>
-          ledgerColl.put(ledgerHash, ledger)
-      }
-      _ <- putMetaData(Bucket(index), ledgerHash)
-      _ <- putMetaData(BucketIndex, index)
-    } yield ()
-
-  /** Retrieve a ledger by state hash, if we still have it. */
-  def get(ledgerHash: Ledger.Hash): KVStoreRead[N, Option[Ledger]] =
-    ledgerColl.read(ledgerHash)
-}
-
-object LedgerStorage {
-
-  /** Keys for different pieces of meta-data stored under a single namespace. */
-  sealed trait MetaKey[V]
-
-  /** Key under which the last written index of the ring buffer is stored. */
-  case object BucketIndex extends MetaKey[Int]
-
-  /** Contents of a ring buffer bucket by index. */
-  case class Bucket(index: Int) extends MetaKey[Ledger.Hash] {
-    assert(index >= 0)
-  }
-
-  implicit val metaKeyEncoder: Encoder[MetaKey[_]] =
-    scodec.codecs.implicits.implicitIntCodec.asEncoder.contramap {
-      case BucketIndex   => -1
-      case Bucket(index) => index
-    }
+    put(ledger.hash, ledger)
 }
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
@@ -1,5 +1,6 @@
 package io.iohk.metronome.checkpointing.service.storage
 
+import cats.implicits._
 import io.iohk.metronome.checkpointing.models.Ledger
 import io.iohk.metronome.storage.{KVRingBuffer, KVCollection, KVStore}
 import scodec.Codec
@@ -38,5 +39,5 @@ class LedgerStorage[N](
     * by going through a block pointing at them directly.
     */
   def put(ledger: Ledger): KVStore[N, Unit] =
-    put(ledger.hash, ledger)
+    put(ledger.hash, ledger).void
 }
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
@@ -1,0 +1,91 @@
+package io.iohk.metronome.checkpointing.service.storage
+
+import cats.implicits._
+import io.iohk.metronome.checkpointing.models.Ledger
+import io.iohk.metronome.storage.{KVCollection, KVStore, KVStoreRead}
+import scodec.{Decoder, Encoder, Codec}
+
+/** Storing the committed and executed checkpoint ledger.
+  *
+  * Strictly speaking the application only needs the committed state,
+  * since it has been signed by the federation and we know it's not
+  * going to be rolled back. Uncommitted state can be kept in memory.
+  *
+  * However we want to support other nodes catching up by:
+  * 1. requesting the latest Commit Q.C., then
+  * 2. requesting the block the Commit Q.C. points at, then
+  * 3. requesting the ledger state the header points at.
+  *
+  * We have to allow some time before we get rid of historical state,
+  * so that it doesn't disappear between step 2 and 3, resulting in
+  * nodes trying and trying to catch up but always missing the beat.
+  *
+  * Therefore we keep a collection of the last N ledgers in a ring buffer.
+  */
+class LedgerStorage[N](
+    ledgerColl: KVCollection[N, Ledger.Hash, Ledger],
+    ledgerMetaNamespace: N,
+    maxHistorySize: Int
+)(implicit codecH: Codec[Ledger.Hash]) {
+  import LedgerStorage._
+  import scodec.codecs.implicits.implicitIntCodec
+
+  private implicit val kvn  = KVStore.instance[N]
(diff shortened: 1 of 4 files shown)


metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
```diff
@@ -0,0 +1,42 @@
...
+/** Storing the committed and executed checkpoint ledger.
+  *
+  * Strictly speaking the application only needs the committed state,
+  * since it has been signed by the federation and we know it's not
+  * going to be rolled back. Uncommitted state can be kept in memory.
+  *
+  * However we want to support other nodes catching up by:
+  * 1. requesting the latest Commit Q.C., then
+  * 2. requesting the block the Commit Q.C. points at, then
+  * 3. requesting the ledger state the header points at.
```
rtkaczyk, line 16: Isn't that the same as latest "committed state"? What more historical state do we want to maintain?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
```diff
@@ -0,0 +1,42 @@
...
+  * We have to allow some time before we get rid of historical state,
+  * so that it doesn't disappear between step 2 and 3, resulting in
+  * nodes trying and trying to catch up but always missing the beat.
+  *
+  * Therefore we keep a collection of the last N ledgers in a ring buffer.
+  */
+class LedgerStorage[N](
+    ledgerColl: KVCollection[N, Ledger.Hash, Ledger],
+    ledgerMetaNamespace: N,
+    maxHistorySize: Int
```
rtkaczyk, line 27: How do we choose this number? Could we not use the committed state as a cutoff point instead?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
```diff
@@ -0,0 +1,42 @@
...
+  * 3. requesting the ledger state the header points at.
+  *
+  * We have to allow some time before we get rid of historical state,
+  * so that it doesn't disappear between step 2 and 3, resulting in
+  * nodes trying and trying to catch up but always missing the beat.
+  *
+  * Therefore we keep a collection of the last N ledgers in a ring buffer.
+  */
+class LedgerStorage[N](
+    ledgerColl: KVCollection[N, Ledger.Hash, Ledger],
```
rtkaczyk, line 25: Would it make sense to use `Block.Hash` as the key instead? It is a 1:1 relationship and we could obtain the parent block's state directly rather than first obtaining the block to read its `postStateHash`. `Ledger.Hash` would then only be used for validation. Of course that only gives us 1 hop in the block path, so I'm not sure if it's really useful.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
rtkaczyk: Do you mean consecutive committed states (ignoring the intermediate uncommited states)?

metronome/storage/src/io/iohk/metronome/storage/KVRingBuffer.scala
```diff
@@ -0,0 +1,74 @@
...
+            coll.delete(oldestKey)
+
+        case None =>
+          coll.put(key, value)
+      }
+      _ <- putMetaData(Bucket(index), key)
+      _ <- putMetaData(BucketIndex, index)
+    } yield ()
+
+  /** Retrieve an item hash hash, if we still have it. */
```
rtkaczyk, line 51: typo: item by hash

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
aakoshh: That is indeed the latest committed state. The only need for keeping any historical state is to facilitate other members trying to catch up. If we completely forget what our state was 5 seconds ago, it might make it difficult for them to query in the window when the state is still available. 

I think we can save the state for any block that we have executed, because every executed block has carried out some effect, so ideally I don't want to execute them again. Not every block has a commit, though, the `BlockExecutor` executes them one by one, resuming from where it left off if it is restarted, since the next commit will implicitly build on any previous one. At that point ideally we should be able to restore the state that belonged to the last executed block, so it's better to save it then not. 

But uncommitted states won't be requested by other federation members to sync to, only committed ones.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
aakoshh: I'm not sure, I guess we just have to find a value that is high enough so we don't see an errors logged. If we have a block every 5 seconds, then maybe save an hours' worth of state. Similarly to block pruning. I think I wanted to keep these two separate so that even if someone doesn't use block pruning, we still don't keep around state forever after it's been committed. 

If I understand the question, you're asking whether we could remove the state as soon as it's committed? Or at the next commit? Again I just want to avoid that if we get a new state every 1 second, then we get 404 every time we try to sync with someone. This fixed size seemed easier to implement than a timestamp based solution, but that's all I really wanted to achieve, have a few minute window to ask for history.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
aakoshh: Great question, I wanted to do this too in the robot example. The reason I didn't is because I figured that if we did that, then there would be no way to validate whether the state returned in the [getState](https://github.com/input-output-hk/metronome/pull/42/files#diff-fcfbaee6667bbaa3ca47408851c4d5b631b40a7143915eb9ed020d267328e3f8R177-R187) response was valid. Now that I'm thinking about it again, it's still possible if we have the block.

So, I want the state at block B, at the moment I retrieve it by `getState(block.postStateHash)` and verify that `response.state.hash == postStateHash`. In the world where state is stored by block key, it would be just `getState(block)` and verifying that `response.state.hash == block.postStateHash`. 

I think I just didn't think it through enough :confused: 

Another drawback could be that if the state doesn't change at all, we still insert it again into the store.

aakoshh: OTOH I just realised that my code assumes that it can delete a state from the collection if it's in the oldest key position, it doesn't consider that the interim positions may refer to it as well :disappointed: 
Storing by block hash wouldn't have that problem.

aakoshh: What I really wanted was not to store state hash in the block, that's what confused me. If not wasn't there, then it would really be impossible to validate a state we fetched.

aakoshh: Fixed it by refcounting. It's up to you what key you want to use. You are right that it could save 1 query in the robot example each time I'm projecting the state from the last executed block hash.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
rtkaczyk: Why would a member try to sync from an old committed state? Perhaps to avoid a race condition we should keep the most recent as well as the previous committed state, but I don't see why we would need more than those 2 🤔

rtkaczyk: For simplicity I wouldn't store the uncommitted state. Executing the blocks should be relatively cheap and we can always think about an optimisation later, though in practice I think that would be a downgrade in performance, as re-executing some blocks upon a restart every once in a while should be cheaper than storing the result of each execution.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
rtkaczyk: Assuming we're storing only committed state there should always be a state to respond with to sync queries. Why would there be a 404? Why would we need hours of historical data?

aakoshh: I mean if we only store the _last_ committed state and it changes quicker than how another node that realises that it's out of sync would fetch it before it's overwritten by the next commit. 

Say Alice and Bob are in sync and they keep producing new commits every 1-5 seconds, and the round timeout is 10 seconds. Charlie sees that it got no valid messages in a round when he times out, and starts syncing with the others. He asks everyone for their last commit quorum certificate, which can take some time if someone is not responding. Then he asks them for the state that corresponds to the last committed block as well -> if at that point everyone the state is no longer available because there was a new commit between Alice and Bob, he has to start again. 

So we don't need hours, just enough for that not to happen.

aakoshh: It could be solved by other means like Guava caching; the ring buffer was just an idea I came up with to make sure we don't end up spinning our wheels with no control over it due to timing issues. Just want it to be a reasonable expectation that if Alice tells Charlie her latest block is X, then Charlie can follow up with the question of the state at that block is without running into data availability issues.

rtkaczyk: Ok, I did not consider querying for QC and the associated state would be separate queries, but yeah, it does make sense.

Btw, I'm not questioning the ring buffer, I really like it :)

rtkaczyk (approved):

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
aakoshh: First let's clarify some terminology. By uncommitted state we can mean two things. 

Say we have these blocks: `B0 <- B1* <- B2 <- B3** <- B4 <- B5` and the last committed block at the moment is `B3`; before that we committed `B1`. 

By _commit_ in this instance I mean that there's an explicit Commit Quorum Certificate for those blocks. A block doesn't necessarily get a Commit Q.C. in a round, but it's also not necessarily discarded: if it has a Prepare Q.C. then it can be extended in next round. Let's say this is how `B2` and `B4` are in the block tree. 

When we validate `B5` we need to project the pre-state it's built on. Similarly if we want to extend it. These projected state are the ones I'd unequivocally call _uncommitted_ and I agree that they should not be stored. 

The other case is `B2`. This one is implicitly committed when `B3` is committed. In theory there can be an arbitrarily long sequence of blocks without a commit, but once it arrives, everything in the branch is committed in one go.

When `B3` is committed `BlockExecutor` will execute blocks `B2` and `B3`. We have bookkeeping for what the last executed block is, even if it happened to be many blocks behind the last commit. I think it makes sense to save this state, because the state change is the de-facto effect of executing a block. 

Arguing whether saving the state or re-executing blocks is cheaper only makes sense in the context of a given application. It could be really expensive and irreversible to execute a block, have real world effects that should only happen once. I think the HotStuff module should not re-execute the same block once it's been successfully done once. 

In theory the application could do its own book-keeping to what it considers the last state and try to retrieve blocks that are built on it when a restart happens, but in our case this sounds more complex than just saving the small state once every few seconds. In the case of ETC, that state won't even change with each block, since there won't be a new checkpoint at every height, so there will only be really new state once per minute.

> Why would a member try to sync from an old committed state? Perhaps to avoid a race condition we should keep the most recent as well as the previous committed state, but I don't see why we would need more than those 2

I don't think you'd deliberately try to sync from "old" committed state, but who knows how quickly something becomes old. You're exactly right, 2 should be enough to avoid the simplest race condition, but some use cases may need a bit more leeway. Maybe "state" isn't as simple as it is for us now, maybe it's a full Merle Patricia Trie that has to be negotiated for what the diff is through multiple rounds of requests.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
aakoshh: Cool; at some point I should add separate tests for it, just made it generic because I wanted to see this working with the Robot example, so it's a bit more representative of what could be an approach that involves state sync.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/storage/LedgerStorage.scala
aakoshh: If there were expensive real world effects for block execution it would probably have to protect itself against re-deliveries, maybe have a nonce built into the commands, because it couldn't be sure that we correctly save the fact that a block has been executed. 

Nevertheless, another reason I thought it should be okay to save the state of blocks like `B2` is because these blocks could have their own Commit Q.C. anyway, it just so happened that they didn't, so the application should be able to handle this many writes, although maybe in a less burst-y fashion. You may be right that performance could become uneven if commits don't happen regularly, in which case we need to tune the timeout.

aakoshh: Another piece of food for thought is that I added [block pruning](https://github.com/input-output-hk/metronome/pull/42/commits/7c4c9cb344ed1ac007fa7d7972279b00469454f4) to the example application based on the last executed block hash, so if the application happened to not save a lot of the state, it may not be able to reconstruct it after a restart, although this would be extreme. And in any case the checkpointing app can have different pruning strategy, one based on what it thinks the last restartable block was.

PR #35, opened 2021-04-26
PM-3063: View synchronisation

aakoshh:
Added a `ViewSynchronizer` component that asks all federation members for their current `Status` and determines the best values to adopt: it picks the highest Prepare and Commit Quorum Certificates, and the median View Number. The former have signatures to prove their validity, but the latter could be games by adversarial actors, hence not using the highest value. Multiple rounds of peers trying to sync with each other and picking the median should make them converge in the end.

TODO:
- [x] Validate signature
- [x] Tests
- [x] Wire into the SyncService
- [ ] ConsensusService to determine when to kick off view synchronisation
- [x] ConsensusService to adopt the new view

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -251,7 +251,10 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
           hotstuff.forensics
         )
 
-      object test extends TestModule
+      object test extends TestModule {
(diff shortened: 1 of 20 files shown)


aakoshh:
Split this up into https://github.com/input-output-hk/metronome/pull/35 and https://github.com/input-output-hk/metronome/pull/36

PR #36, opened 2021-04-28
PM-3063: View synchronisation (Part 1)

aakoshh:
Added a ViewSynchronizer component that asks all federation members for their current Status and determines the best values to adopt: it picks the highest Prepare and Commit Quorum Certificates, and the median View Number. The former have signatures to prove their validity, but the latter could be gamed by adversarial actors, hence not using the highest value. Multiple rounds of peers trying to sync with each other and picking the median should make them converge in the end.

The integration with the services is in https://github.com/input-output-hk/metronome/pull/37

### DIFF
--- a/metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/ViewNumber.scala
+++ b/metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/ViewNumber.scala
@@ -1,6 +1,7 @@
 package io.iohk.metronome.hotstuff.consensus
 
 import io.iohk.metronome.core.Tagger
+import cats.kernel.Order
 
 object ViewNumber extends Tagger[Long] {
   implicit class Ops(val vn: ViewNumber) extends AnyVal {
@@ -10,4 +11,7 @@ object ViewNumber extends Tagger[Long] {
 
   implicit val ord: Ordering[ViewNumber] =
     Ordering.by(identity[Long])
+
+  implicit val order: Order[ViewNumber] =
+    Order.fromOrdering(ord)
 }
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
@@ -3,7 +3,7 @@ package io.iohk.metronome.hotstuff.service.sync
 import cats._
 import cats.implicits._
 import cats.effect.{Timer, Sync}
-import cats.data.NonEmptyVector
+import cats.data.NonEmptySeq
 import io.iohk.metronome.core.Validated
 import io.iohk.metronome.hotstuff.consensus.{Federation, ViewNumber}
 import io.iohk.metronome.hotstuff.consensus.basic.{
@@ -45,7 +45,7 @@ class ViewSynchronizer[F[_]: Sync: Timer: Parallel, A <: Agreement: Signing](
           .statusPoll(federation.publicKeys -> maybeStatuses)
           .as(maybeStatuses.flatten)
       }
-      .map(NonEmptyVector.fromVector)
+      .map(NonEmptySeq.fromSeq)
       .flatMap {
         case Some(statuses) if statuses.size >= federation.quorumSize =>
           aggregateStatus(statuses).pure[F]
@@ -133,7 +133,7 @@ object ViewSynchronizer {
   type GetStatus[F[_], A <: Agreement] = A#PKey => F[Option[Status[A]]]
 
   def aggregateStatus[A <: Agreement](
-      statuses: NonEmptyVector[Status[A]]
+      statuses: NonEmptySeq[Status[A]]
   ): Status[A] = {
     val prepareQC = statuses.map(_.prepareQC).maximumBy(_.viewNumber)
     val commitQC  = statuses.map(_.commitQC).maximumBy(_.viewNumber)
@@ -146,6 +146,6 @@ object ViewSynchronizer {
     )
   }
 
-  def median[T: Order](xs: NonEmptyVector[T]): T =
+  def median[T: Order](xs: NonEmptySeq[T]): T =
     xs.sorted.getUnsafe((xs.size / 2).toInt)
 }
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
@@ -65,8 +65,8 @@ class ViewSynchronizer[F[_]: Sync: Timer: Parallel, A <: Agreement: Signing](
 
       case Some(status) =>
         validate(from, status) match {
-          case Left(error) =>
-            tracers.invalidStatus(status, error).as(none)
+          case Left((error, hint)) =>
+            tracers.invalidStatus(status, error, hint).as(none)
           case Right(valid) =>
             valid.some.pure[F]
         }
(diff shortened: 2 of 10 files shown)


metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/ViewNumber.scala
```diff
@@ -10,4 +11,7 @@ object ViewNumber extends Tagger[Long] {
 
   implicit val ord: Ordering[ViewNumber] =
     Ordering.by(identity[Long])
+
+  implicit val order: Order[ViewNumber] =
+    Order.fromOrdering(ord)
```
rtkaczyk, line 16: Just curious, what does `Order` gives us, that `Ordering` does not?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
```diff
@@ -0,0 +1,151 @@
...
+
+object ViewSynchronizer {
+
+  /** Extra textual description for errors. */
+  type Hint = String
+
+  /** Send a network request to get the status of a replica. */
+  type GetStatus[F[_], A <: Agreement] = A#PKey => F[Option[Status[A]]]
+
+  def aggregateStatus[A <: Agreement](
```
rtkaczyk, line 135: This could use a similar explanation as in the PR description

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
```diff
@@ -0,0 +1,151 @@
...
+    val viewNumber =
+      math.max(median(statuses.map(_.viewNumber)), prepareQC.viewNumber)
+    Status(
+      viewNumber = ViewNumber(viewNumber),
+      prepareQC = prepareQC,
+      commitQC = commitQC
+    )
+  }
+
+  def median[T: Order](xs: NonEmptySeq[T]): T =
```
rtkaczyk, line 149: This is not technically a median, as `median([0, 2]) === 1`. I think this fits the definition of a [medoid](https://en.wikipedia.org/wiki/Medoid). I'm fine with this, but we should perhaps make it clear in a comment?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizerProps.scala
```diff
@@ -0,0 +1,374 @@
...
+        }
+        groupSig = mockSigning.combine(partialSigs)
+      } yield QuorumCertificate[TestAgreement](
+        Phase.Prepare,
+        viewNumber,
+        blockHash,
+        groupSig
+      )
+
+    def genCommitQC(qc: QuorumCertificate[TestAgreement]) =
```
rtkaczyk, line 144: `genCommitQC` and `genPrepareQC` are very similar. Could we DRY them?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizerProps.scala
```diff
@@ -0,0 +1,374 @@
...
+    for {
+      m   <- arbitrary[Int].map(_.toLong)
+      l   <- Gen.posNum[Int]
+      h   <- Gen.oneOf(l, l + 1)
+      ls  <- Gen.listOfN(l, Gen.posNum[Int].map(m - _))
+      hs  <- Gen.listOfN(l, Gen.posNum[Int].map(m + _))
+      rnd <- arbitrary[Int].map(new Random(_))
+    } yield (m, rnd.shuffle(ls ++ Seq(m) ++ hs))
+  ) { case (m, xs) =>
+    m == ViewSynchronizer.median(NonEmptySeq.fromSeqUnsafe(xs))
```
rtkaczyk, line 338: I think you meant to generate `hs` using `h`, otherwise we're only testing lists of odd lengths. This would fail however for `h == l + 1`. We need `h ∈ [l - 1, l]` for the current impl.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
rtkaczyk: What if used a _mode_ instead? We could calculate all modes and take a "median" of those, if in an extreme case all compared view number are different.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
```diff
@@ -0,0 +1,151 @@
...
+  ): Either[
+    (ProtocolError.InvalidQuorumCertificate[A], ViewSynchronizer.Hint),
+    Validated[Status[A]]
+  ] =
+    for {
+      _ <- validateQC(from, status.prepareQC)(
+        checkPhase(Phase.Prepare),
+        checkSignature,
+        checkVisible(status),
+        qc =>
```
rtkaczyk, line 87: Minor, subjective: for improved aesthetics I'd give it a name, even if used only once

metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/ViewNumber.scala
aakoshh: Had to comment it to to see what fails...

It's [maximumBy](https://typelevel.org/cats/api/cats/NonEmptyReducible.html#maximumBy[A,B](fa:F[A])(f:A=%3EB)(implicitevidence$3:cats.Order[B]):A) which I used [here](https://github.com/input-output-hk/metronome/pull/36/files#diff-e5099455fe288ae75f26d96087e70bb166397f06b80c0ba11201605a3ed9cbf3R138-R139) on `NonEmptySeq` instances.

rtkaczyk (commented):
Conceptually I like it! Only picking on minor details. Ready to approve once these are resolved

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: You are right, I'll add it, thanks.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: Thanks, I didn't know about medoids :)

I named _median_ this because I thought this is the most intuitive name that people will immediately have a pretty good understand of what we are doing and why. I know it doesn't match the traditional definition, I was thinking about it more as an [ordinal value](https://stats.stackexchange.com/questions/11219/median-value-on-ordinal-scales). I thought taking the average of the middle numbers would land you in a no-mans-land, rather than on a view number occupied by at least somebody. You are right, I'll make a comment about it.

No idea whether _mode_ would be better. If as you say the federation is spread out over many view numbers, adversarial nodes could team up to present random maximums to different honest parties, keeping everyone scattered. Like if Alice, Bob and Charlie are honest and have different view numbers, say, 10, 20 and 30, Eve and Edward could tell Alice that they both have 40, to Bob that they are both 50, to Charlie they are both 60. 

Are you're suggesting then that all the others' single-member modes would be mixed in with the modes, so one would pick from [10,20,30,50], the other from [10,20,30,60]? How do you take the height of the mode into account in this case? 

I guess a median is also susceptible to such attack when the network is completely out of sync. Mode in that case could be better if there's at least `f+1` people in the same view.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizerProps.scala
aakoshh: Certainly! No idea why I left it like this.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizerProps.scala
aakoshh: Thanks, you have eagle eyes!

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: Sorry, I meant categorical values, not ordinal.

rtkaczyk: Yeah, it wasn't a well thought-out suggestion. Both approaches could be attacked as you say.

But I'm thinking of different approach now. What if we sync to at most `max(prepareQC.viewNumber, commitQC.viewNumber + 1)` or maybe just `lastQC.viewNumber + 1`? This could require an honest node to roll back its view number, but only in the case that the preceding rounds were not successful at producing a QC. I realise that this would affect leader selection, but once the honest nodes are in sync, they could continue even if the selected leader is dishonest. Would that make sense?

I'm not sure, seems risky. Looking for another approach as it seems to me any statistical method could be susceptible to attack.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: Done, thanks, it was ugly indeed. Left it under the prepare QC checks, rather than a new "status level" block, since there's just the `InvalidQuorumCertificate` error available for this; have to put the blame on one of the QCs.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizerProps.scala
aakoshh: Done, good catch.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizerProps.scala
aakoshh: Done, thanks!

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: Copied descriptions and added comments to median as well.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: Actually I am doing a [+1](https://github.com/input-output-hk/metronome/blob/pm-3323-example-app-integ-test/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala#L222) now, trying to front run the view I hope everyone else will be finishing soon.

aakoshh: Maybe it's worth creating a separate ticket to do some research on this topic. I remember reading that Libra does regular gossip about status, but that changes the complexity of the protocol. We can find out what they do exactly to better understand the landscape.

rtkaczyk: Yes, I agree about the separate ticket. This definitely warrants more research, but it's unlikely to cause issues for now, unless more than 1 actors collude to disrupt syncing.

rtkaczyk (approved):
:+1:

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/ViewSynchronizer.scala
aakoshh: Cool, created https://jira.iohk.io/browse/PM-3346

PR #37, opened 2021-04-28
PM-3063: View synchronisation (Part 2)

aakoshh:
Wires the `ViewSynchronizer` added in https://github.com/input-output-hk/metronome/pull/36 with the `SyncService` and the `ConsensusService.`

The flow is that the `ConsensusService` counts how many messages it receives in each round that are in/out of sync with its view number and uses it to make a decision whether a full view synchronisation is needed in case there is a timeout. If it's needed, a request is added to the `SyncPipe` for the `SyncService`.

The `SyncService` has two sync modes,block or view. In block mode it downloads any missing ancestors of new `Prepare` messages. If view mode is requested, it cancels all outstanding block synchronisation, and asks the `ViewSynchronizer` to poll the federation for the `Status`. Until that's done, any further prepare and view sync requests are dropped. When the `Status` is available, it downloads the block contained in the `commitQC`, clears out the block storage, and inserts the new block as the root of the block tree. Before that it tells the application layer (i.e. the Checkpointing Service) to synchronise the ledger. Finally it sends a response in the `SyncPipe` for the `ConsensusService`. 

When the `ConsensusService` sees the newly adopted `Status` it applies it on the `ProtocolState` and triggers a transition to the next view.

I added a `RootBlockHash` key to `ViewStateStorage` so that we can track what was the oldest block in storage should be, which is the basis of what can be cleared out when we execute a "jump" to a new committed state by skipping the interim block history.

Added a component diagram to the [README](https://github.com/input-output-hk/metronome/blob/ee08e7a4e74fd379005a8abab2bd9d9bd6b6b4a7/README.md#architecture)


metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
```diff
@@ -158,49 +169,174 @@ class SyncService[F[_]: Sync, N, A <: Agreement](
...
+    // If the same leader is sending us newer proposals, we can ignore the
+    // previous pepared blocks - they are either part of the new Q.C.,
+    // in which case they don't need to be validated, or they have not
+    // gathered enough votes, and been superseded by a new proposal.
+    blockSync.fiberMap.cancelQueue(sender) >>
+      blockSync.fiberMap
+        .submit(sender) {
+          for {
+            _       <- blockSync.synchronizer.sync(sender, prepare.highQC)
+            isValid <- appService.validateBlock(prepare.block)
```
rtkaczyk, line 244: Is this the right thing to do? In order to validate block contents we will need the state leading up to it and, while the `appService` could maintain such state, we cannot guarantee that the block referred in highQC has been executed yet. Shouldn't we instead only rely on `safeNode` validation up to this point and defer interpreter-assisted block contents validation until execution on commit?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
```diff
@@ -34,20 +35,26 @@ import scala.util.control.NonFatal
   *
   * It handles the `consensus.basic.Message` events coming from the network.
   */
-class ConsensusService[F[_]: Timer: Concurrent, N, A <: Agreement: Block](
+class ConsensusService[F[
+    _
+]: Timer: Concurrent, N, A <: Agreement: Block: Signing](
```
rtkaczyk, line 40: This is the first time I see scalafmt fail so spectacularly. Fortunately it can be prettified to something like:

```scala
class ConsensusService[
    F[_]: Timer: Concurrent,
    N,
    A <: Agreement: Block: Signing
](
```

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: This validation should not require execution, only checking by way of modeling the change I suppose, ie. applying it on the ledger, which we can do by extension from the last executed ledger. I reckon we should only vote on blocks which have valid content. At least that's how I understood @ranvirranaiitb paper on Advocate. That's why we pass the ledger in the validation request, because the interpreter should make its decisions based on that.

Otherwise, if we could commit and execute an invalid block, the BFT proof we can produce over it, which goes in the checkpoint certificate, would be worthless to any outsider: yeah we agreed to checkpoint block B, but it may not even exist, and we can change it to B' any time.

rtkaczyk: > ie. applying it on the ledger, which we can do by extension from the last executed ledger

What I'm saying is that we may not the previous the previous ledger. The parent of the block to be validated may not have been executed

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
```diff
@@ -461,19 +517,38 @@ object ConsensusService {
     def empty[A <: Agreement] = MessageStash[A](Map.empty)
   }
 
+  /** Count the number of messages received from others in a round,
+    * to determine whether we're out of sync or not in case of a timeout.
+    */
+  case class MessageCounter(
+      past: Int,
```
rtkaczyk, line 524: I understand the desire for symmetry, but it doesn't look like `past` is needed at all

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Indeed, and that is perfectly normal. 

Say we have a block tree like this, where `C*` are Committed blocks, `PC*` is PreCommitted, and `P*` is Prepared, with  the view number at the end.
```
C0 - C1 - PC4 - P5 - P6
             \ 
               P2 - P3
```
It is possible that in a round we Prepare a block but don't get to the PreCommit phase, and the next leader chooses to extend another earlier Prepare Q.C. According to the paper, it is legal for the leader to just extend whatever his own highest Prepare Q.C. is, doesn't necessarily have to wait for NewView messages to arrive. Nobody can tell whether they did so anyway, all they can do is not vote on it if it's not a safe extension. 

So it's possible that we get P2, but it doesn't get enough votes to get locked; nevertheless it's the highest QC, so the next leader extends it with P3, which also doesn't get to the end of the voting. Maybe the next leader didn't get these, so they extend C1 again with P4, and this one gets enough votes to become PC4, and then gets extended with P5 and P6. 

At this point neither of the P* blocks are executed yet because they aren't committed. Yet you can decide whether P6 is valid by doing: `isValid(block = P6, ledger = getLedger(C1.postStateHash).update(PC4).update(P5))`. You know that PC4 and P5 are valid, because they have Quorum Certificates to back that up (assuming the majority of the federation is honest), so when you received `Prepare(block = P6, highQC = QC(blockHash = P5.hash))` you knew that all you have to check is whether P6 is a valid block, you can trust anything up to P5. 

We can do these checks without executing blocks and saving the state to storage, and carrying out other side effects, doing it all in memory. This is somewhat similar to what Mantis is doing with the `TentativeBranchManager` and `TentativeCompleteTree` stuff: there you may discover that you have a better branch, in which case you have to roll back to where the fork point was, and re-apply all blocks on the longer branch. What's called a stable block there, is akin to our commit state, that will not be rolled back. 

In the robot example I did this in [RobotService.projectState](https://github.com/input-output-hk/metronome/pull/42/files#diff-fcfbaee6667bbaa3ca47408851c4d5b631b40a7143915eb9ed020d267328e3f8R63-R93) where for simplicity, the process always starts from the `lastExecutedBlockHash`, whatever that might be, retrieves all blocks between that and the latest proposal and applies them in order. I expect that in checkpointing we can spend more time on this to have a bit more finesse and maintain some sort of in-memory tree of the tentative blocks / ledger states that need to be passed to the interpreter, and also in order that we can dynamically filter the mempool based on what we see on the given branch; the principle is the same though.

But it's essential that we don't commit an invalid block. In HotStuff, a classical BFT, that would mean the federation agreed permanently to that block, there's no going back, so it better be valid! The signed Commit Q.C. is what separates this solution from Ouroboros BFT: we can show someone that the block the Q.C. is about is legit, and we will not change our minds about it.

aakoshh: If you're wonder: why don't we do the check just _before_ voting Commit, well, firstly, that's not how the paper describes the algorithm, secondly, that would imply we already have a PreCommit Q.C. for a potentially invalid block, which would lock it in. Then why don't we validate after the arrival of the Prepare Q.C.? Similar reason: in Prepare we get an extension of a previous Prepare Q.C.; if we can't trust that its' correct, we have to re-validate everything back to whatever we validated last time, which is more difficult to keep track of.

aakoshh: The validation arrangements are on page 9 of the Advocate paper:
image

As you can see, the "BFT module", which is our `hotstuff.consensus.basic.ProtocolState`, doesn't "receive" the message until it goes through validation.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
aakoshh: Wow! I tried many times but without the final linebreak before the `]` it just put it back the way it was. Thanks!

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
aakoshh: Yeah, syncing the view number is not something that's covered in the HotStuff paper, so I just came up with something that seemed plausible. Initially I thought we could keep tabs of what view numbers we receive in messages, try to figure it out from those, but then decided to start with this simpler model of just polling the others if we time out. This is rather experimental and subject to change. 

I thought I'd gather the statistic anyway so that at least we can [trace them](https://github.com/input-output-hk/metronome/pull/37/files#diff-6304157ee8ddbbd937b7fb96c7cc976949ff9833f037ee0032780f8d32e3e7b3R232) and see what's happening. While watching the terminals in the robot example trying to get in sync, I actually [modified](https://github.com/input-output-hk/metronome/pull/42/files#diff-6304157ee8ddbbd937b7fb96c7cc976949ff9833f037ee0032780f8d32e3e7b3L259-R284) this logic to make use of the past.

What I realised was that the node with the highest view number can get isolated. Say we start nodes Alice, Bob and Charlie in a staggered fashion, in that order, and by the time Charlie starts their view numbers are 100, 10, 1, respectively. Bob and Charlie will see that they sometimes get messages from the future, when an older node is leading the round, so they initiate view synchronisation. They, however, pick the median view number, which is Bob's, so they converge after a while, but Alice only gets messages from the past, so she never starts the syncing process. 

So I thought: it should be okay for Alice to wind back her view number, if she sees that she's only getting past messages while timing out. 

These heuristics are not infallible, for example if we had 2 groups with identical view numbers within the group, but different between them, they could receive present messages (multiple nodes can think they are leading, one in each group), but I don't have a better idea at the moment, I'm just watching how long it takes to converge. 

I chose the median because I didn't want to trust the highest view, which could be just one Byzantine node trying to lie to us to isolate us from the herd. I thought if the majority of the federation is in sync, then the median is the right number to use for a new node.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
rtkaczyk: > isValid(block = P6, ledger = getLedger(C1.postStateHash).update(PC4).update(P5))

Ok, sure we can do it like that. But why wouldn't we just fully execute the blocks instead and save the intermediate states thus avoiding repeated `update`s on the same block? Granted `update` is cheap in our application, but I'm thinking about the generalised case. Image running Smart Contracts on this chain.

> But it's essential that we don't commit an invalid block.

Totally! But we cannot fully validate a block until we execute it, because of `postStateHash`. Yes, we can use the `update` trick again, but why pretend we're not executing a block when we are, in fact, doing it. The ledger update function is the essence of block execution and, at least to me, they're almost synonymous.

In this design the difference between executing a block and calling the `update` function on it is basically how long we keep the resulting state in memory. So what I'm saying essentially boils down to a question:

Do we have to execute the blocks in committed batches rather than immediately upon receiving them in `prepare`?

rtkaczyk: I would argue that in the _general_ case full block validation may depend on block execution. This is true for most of the practical blockchains that I know of.

aakoshh: Depends on what you mean by execution I suppose. 

The paper has scant information on this:
> Additionally, in order to incrementally execute a committed log of commands, the replica maintains the highest node whose branch has been executed.

> DECIDE phase
>   as a replica
>     wait for Commit QC
>     execute new commands through m.justify.node, respond to clients

So the consensus is about what commands to execute. As long as they fail the same way and every client sees the same results, good.

But if we consider our checkpointing use case, where we want to use the proof-of-agreement to pick which fork to execute, I maintain that it would make no sense to commit the checkpointing federation to a block that either doesn't exist or it is invalid in the PoW side. 

I think in the case of checkpointing, such use the `update` function to tell the interpreter what context to validate a block in is reasonable. 

Take this example, say we have missed producing a checkpoint and now we have two competing forks in the PoW chain:
```
                     B5' - B6'* - B7' - B8' - B9'*
                   /
B0 - B1 - B3** - B4 
                   \ 
                     B5 - B6* - B7 - B8 - B9* - B10
```
Say we want to checkpoint every 3rd block, and the last one we managed to commit and produce a Certificate for was B3. Some nodes went with the upper fork, some with the lower fork. 

When a leader whose PoW node follows the upper node is about to propose a block to the Checkpointing Federation, they propose `B9'`, but in their ledger the last checkpoint is `B6'`, so every federation member gets this message: 
```
ValidateBlockBody(
  blockBody = BlockBody(transactions = List(CheckpointCandidate(B9'))),
  ledger = Ledger(maybeLastCheckpoint = Some(CheckpointCandidate(B6')))
)
```
Validation could go like this: 
0. On each node, the last checkpoint certificate is for B3, so they have to establish whether B6' actually extends B3. 
1. Those on the lower fork have two choices:
1.a) They accept that B6' is a valid extension of B3, and B9' is a valid extension of B6', and although they think B10 is a longer chain, they accept that this other fork is valid too, respond `true`, and wait to see what happens. 
1.b) They see that would be a valid checkpoint for the upper fork, but they think that fork is worse than what they have, so they vote `false`. 

If B9' gets elected, everyone has to adopt the upper fork. If not, maybe the next leader will propose B9 instead. In either cases, the Certificate for the next checkpoint only gets made in the execution phase, and it would be bad if invalid certificates were circulating.

aakoshh: cc @dzajkowski :point_up:

aakoshh: > but why pretend we're not executing a block when we are, in fact, doing it. The ledger update function is the essence of block execution and, at least to me, they're almost synonymous.

Right, that's indeed where we see things differently. The result of execution in the checkpointing application is not just the persisting of the new ledger state, but the construction and publication of a Checkpoint Certificate which will have an irreversible effect on the PoW chain. Once the certificate is out, nobody can legally build on a block that doesn't extend the block in it. 

In the robot example, the difference between validation and execution would be that in validation we project in memory where the robot would be if we carried out the command, check if it's a valid state, and vote for it if it is, while the execution would be for actually telling the robot to do the work. Imagine if the robot is carrying out an operation on a live patient in a hospital, moving the scalpel left and right while the controlling doctors try to come to consensus of where to cut. Or if it's a bomb disarming robot, where experts decide which wire to snap :bomb: :boom: There are no second chances there.

aakoshh: > But why wouldn't we just fully execute the blocks instead and save the intermediate states thus avoiding repeated updates on the same block? Granted update is cheap in our application, but I'm thinking about the generalised case.

From the perspective of the `SyncService`, we don't know if the `ApplicationService` saves the immediate states, such optimisations are indeed within its jurisdiction. If it's expensive to validate state, sure, cache it, save it. But then, as discussed, there can be some extra stuff happening in the "real" execution.

> Imagine running Smart Contracts on this chain.

That's a good point! In the previous example with the forks, it may be too expensive for the PoW node on the lower fork to roll back to B4, then forward to B9' just to check if its a valid block. In that case it can just decide to vote `false`, or not vote at all, it's up to the PoW interpreter.

rtkaczyk: > DECIDE phase
> as a replica
> wait for Commit QC
> execute new commands through m.justify.node, respond to clients

I would interpret that as describing the situation for an external observer. Only on Commit would the state be recorded as committed and communicated to the client, but nothing prohibits us from calculating that state beforehand. The external observer will still see the most recent state as the last committed state via any queries she may send or receive, so from her point of view the blocks are executed only upon Commit.

This would be similar to how Ethereum-like chains work. A block gets executed with a value transfer to my account, but the wallet that I use tells me my balance hasn't changed. Only after an additional can I utilise those funds. In HotStuff we're in a preferable position because such consensus is unfalsifiable rather than resulting from some external probabilistic heuristic.

In our case we send the CheckpointCertificate to the Interpreter only on Commit, but internally the blocks have been executed earlier. It's a bit more complicated in our case because the Interpreter sees some intermediate state, but that's what we get when make our service depend on the logic of our client :wink:. Importantly though, no CheckpointCertificate would be sent until the corresponding state becomes committed. I believe that invalidates your example, because after B9 or B9' gets committed there could be no forking from that.

aakoshh: It's imperative that we deny even the possibility of a Byzantine federation member to produce a Checkpoint Certificate. If the Commit Q.C. was at hand before the block is validated, they could send it to the PoW network and nobody could say anything about it. 

My point in the example was that we should not commit to B9, and then ask the PoW node whether it's a valid block, because if it's not, we are screwed.

aakoshh: Even Ethereum would not execute an invalid block. So while your balance may not be final, at least it shows promise.

 Prism, on the other hand, will put the block on the chain and execute only once it's stable.

rtkaczyk: I think our difference stems from the fact that we are interpreting _execution_ differently. I think of executing blocks in a blockchain, whereas you think of executing commands as stated in the paper. I think we should dissociate these concepts.

I would postulate the interpretation in which a _command_ is the CheckpointCertificate (or lack thereof) and we _execute a command_ it by sending it to the Interpreter. Block transaction OTOH are not commands, they are pieces of data leading up to formation of a command. When we _execute blocks_ we update the ledger from which we will derive the command.

That would be also consistent with Ethereum-like blockchains, where executing blocks means evolving the ledger, executing command is the act of block confirmation. 

I think you may also be misled by initial question suggesting validation upon execution. I did not mean validation after commit and neither do I suggest anymore that we should validate at a later point. Sorry If I had not made that clear. Instead what I'm proposing is that at this point we already update the ledger and save the resulting state in memory. So to be in line with what I said above I would propose:
* upon `prepare` we call `appService.executeBlock(prepare.block)` which validates the block (using a previously calculated state), updates and saves the ledger in memory
* upon `Commit` we call `appService.commit(commitQC)` (or `executeCommand` if we want to stick to the paper), which looks up the corresponding state, then builds and sends the `CheckpointCertificate`

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
rtkaczyk: Thanks for the explanation

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: I'm glad we are on the same page, I think I've been saying the same thing, that we consider execution to be different. I definitely said that it's up to the ApplicationService implementation whether it saves the uncommitted state to memory or persistent storage, as long as it is also prepared for clearing it out, and not losing the committed one. This is what I suggested we could spend more time on, rather than how I did it with the robot example.

What you are suggesting are mere renamings though: what is currently validateBlock, you would call executeBlock, and what is now execute would be commit.

I don't see a need for this rename. Then current naming is line with both the HotStuff and the Advocate paper, there's no need to confirm to what other block chains call execution IMO. 

There is a difference between the last committed block and the last executed one, though.

aakoshh: One option for storing the "state tree" in memory is to generalise the `BlockStorage` a little bit to not be tied to `Aggrement`, and stick the state into it, keyed by `Block.Hash`. Then use the `deleteNonDescendants` method to clear out orphan branches when a block is finally executed. Can even be persistent but in memory is better, if, as we discussed, there are lots of empty blocks.

aakoshh: I didn't want "command" to appear in our application. I thought "command" in the HotStuff paper is an abstract term for the what is usually the block contents. What's usually called a block chain, the HotStuff paper called node tree. So I thought I'd use a more familiar terms and just call them blocks, and that's what we execute. The paper also doesn't talk about state, because in its abstract form, the consensus is about which blocks/commands to execute.

Also wanted to mention that execution doesn't necessarily mean evolving a ledger and persisting the results. For example Hyperledger does things differently: their scheme is called execute-order-validate. The client first asks a couple of nodes to execute a transaction, to know the outcome and get endorsements. The result contains the key level changes (writes, deletes, version numbers), but doesn't have any effect on the databases. Then the ordering service sorts the transactions arbitrarily, and finally validation checks whether the transactions in their final positions have any conflicts when they are finally attempted to be written (committed) to the database; conflicting ones are marked as invalid and have no effect. So in this instance, "validate" has the modification and "execution" is the tentative one. 

Prism has delayed execution. When a transaction is added to the block, it just does preliminary validation on balances, but doesn't execute smart contracts for example. That comes only when a block is confirmed (committed), which is when they execute the effects and publish the post state hash to an ancestor block. 

So I reckon it's difficult to point to any single canonical meaning of execution. That's why I'd like to stick to the HotStuff paper nomenclature, and the Advocate diagram I pasted above. Those are the most relevant to the project, that's what we are implementing.

rtkaczyk: Firstly, naming is important. I'd like our names to convey what we're doing rather than explain it all in comments. Now if I see that there's a `validateBlock` which executes transactions and updates the ledger and there's also an `executeBlock` which does not do that then I'm confused. And it requires some mental gymnastics, because I think we can agree what "executing a transaction" means, so if we want to execute transactions in the context of their carrying block it's hard not to call that "executing a block".

This is not in conflict with HotStuff paper, all it says is that on commit the committed branch is executed. It shouldn't matter if the blocks were executed before. All that matters in that on commit that branch becomes de facto executed, while any other potentially executed branches will eventually be discarded. HotStuff paper does not mention full validation of blocks so it doesn't have to consider what happens before commit. But since we cannot allow invalid state to be committed then we must produce the state before committing it.

Secondly, this is not just about names, since what I'm suggesting undermines the necessity for `BlockExecutor`, at least in its current form.

aakoshh: I also agree that names are important, but I think these names reflect what we want an abstract `ApplicationService` to do, not what we know a specific implementation we know about has to do. From the consensus perspective, we want the block to be validated, regardless of what needs to happen for that. 

> I think we can agree what "executing a transaction" means

Apparently not :) You could say that in the `checkpointing` module you want to use "execution" to mean applying transactions on the ledger, assuming that you can always unapply them, but this is not a general notion. A blockchain can usually do this, because it's a database that doesn't affect the world, but in our case we also have the "respond to clients" concept (push checkpoint certificate), which we cannot undo. I consider that to be part of execution.

> This is not in conflict with HotStuff paper, all it says is that on commit the committed branch is executed.

This is why there's an [ExecuteBlocks](https://github.com/input-output-hk/metronome/blob/develop/metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/basic/Effect.scala#L65-L68) effect that captures precisely this notion.

> It shouldn't matter if the blocks were executed before. All that matters is that on commit that branch becomes de facto executed, while any other potentially executed branches will eventually be discarded

I don't think that's in the spirit of the paper. The paper says:
> Upon receiving a decide message, a replica considers the proposal embodied in the commitQC a committed decision, and executes the commands in the committed branch. 
> in order to incrementally execute a committed log of commands, the replica maintains the highest node whose branch has been executed

I think the only interpretation of that is we execute blocks which are committed, incrementally, picking up from the last block that has been done before. If it considered multiple branches to have potentially already been executed, and orphaned branches' execution results be discarded, it wouldn't be enough to have a singular "highest executed node" variable in bookkeeping. Using "execution" for anything else in the `hotstuff` module would be confusing.

> what I'm suggesting undermines the necessity for BlockExecutor, at least in its current form.

Executing asynchronously was a design choice that seemed to make sense at the time. The paper doesn't specify whether execution must happen synchronously before the new-view is sent. I thought it didn't, and just enqueued them for the `BlockExecutor` to do so, thinking that it might take some time to execute, and consensus shouldn't be held up unnecessarily. 

This could indeed be done in different ways, we could wait for the execution to be confirmed before moving to the new view. It might make handling `CreateBlock` easier to do so. There's also a slight discrepancy that when we try to sync state, we report some block to be committed but it may not actually have been executed yet, so its state may not be available for fetching right away.

I thought since HotStuff consensus has some high synchrony assumptions, in order to not risk the view number to fall out of sync, it's better to let the application carry out blocks on its own time. Otherwise there are more questions: should we wait until we sent the certificate to the PoW node, or until it has created a checkpoint block and gossiped it around? What if it's not available? What if it sometimes takes longer to carry out an effect than a HotStuff timeout?

aakoshh: > But since we cannot allow invalid state to be committed then we must produce the state before committing it.

Maybe it would help to think about different problems that a classical BFT can be applied to, not just blockchain. For example what if we just want state machine replication on an RDBMS database. 

Say you have client sending these commands: `delete from foo`, `insert into foo (amount) values (1)`, `insert into foo (amount) values (2)`, `select sum(amount) from foo`. The results depend on what order we are carrying out the modifications in, which is what consensus is about.

Validation could simply mean parsing the query, checking that the user the commands are running under are not trying to access a table they are not supposed to (e.g. `drop users`). There's no need to actually execute the query to validate the command, and indeed it may not be possible to roll back the blocks if we executed them. Maybe you could implement commit and rollback on an actual RDBMS, but what if this is just some in-memory mutable storage, perhaps memcached, that doesn't support transactions? 

Maybe there's no need to validate anything, because all commands are equally valid, just setting values in a generic KV store? Would you still call it execution? Isn't `def validateBlock(...) = true` more intuitive in that case then `def executeBlock(...) = true` with `def checkpointBlock(block: KVQuery) = <talk to memcached>`?

rtkaczyk: > Say you have client sending these commands: delete from foo, insert into foo (amount) values (1), insert into foo (amount) values (2), select sum(amount) from foo. The results depend on what order we are carrying out the modifications in, which is what consensus is about.

That's a great example. It actually illustrates both your (paper's) approach as well as mine, as this application could have 2 different sets of requirements:
1. it could simply require a consensus on the order of commands letting the RDBMS resolve any conflicts and invalid input. This fits your proposal where we execute the commands only after commit.
1. it could require that a sequence of valid commands that produce a valid state is agreed upon. In this case the commands have to be executed prior to the commit and the RDBMS has to support a rollback.

So I still stick with my proposal because:
- The latter case fits our primary application
- It also fits most of blockchain-like applications I can think of including you Robot example, where you first apply valid moves and check that the resulting robot coordinates are valid - this is done prior to commit
- It's consistent with what executing a block/tx means in both Mantis and Midnight (it doesn't finalise the state) 
- I don't want to redefine the English language. While we define "execute" to also mean finalisation of state, it just seems natural to me to talk about applying changes to state as executing them. It seems it's also natural to you as that's how you describe what [validateBlock does in your Robot example](https://github.com/input-output-hk/metronome/blob/9e47f5931221da18869a454cb892cbd22a4129fa/metronome/examples/src/io/iohk/metronome/examples/robot/service/RobotService.scala#L125)

However, there might be a way to reconcile both approaches by using more general names:
- `applyBlock` - called before commit. Either performs simple validations (case 1.) or applies and validates state changes (case 2.)
- `commit` - called upon commit. Either executes the blocks (case 1.) or expects and looks up pre-executed state (case 2.)

Now, I realise that a consensus may be hard to reach between us on this matter. We are not a BFT federation after all 😆. I feel that carrying on this discussion may be counter-productive, so if my arguments do not convince you, please know that I'm fine with accepting your vision, especially in light of all the thought and great work you have put into this project.


There is however the second part of my proposal which consists of:
- before commit: applying the blocks to the state and storing the resulting state in memory
- during decide: expecting the state to be precomputed

This approach will probably skip `BlockExecutor` but will also have to validate (update ledger) during block synchronization. Are you ok with me continuing this approach and possibly discussing further on my PR for PM-3108?

aakoshh: Thank you for your continued deep thoughts on this, it's a fun discussion for sure :smiley: 

> It's consistent with what executing a block/tx means in both Mantis and our other project (it doesn't finalise the state)

I'd like to point out that these two will use different checkpointing mechanism as well as different blockchain consensus. Mantis executes the blocks as it comes, Advocate does not, it just validates, only executes later. Validation in Advocate does not require execution, only checking that blocks exist and they are formally valid, also that they extend the previous checkpoint. Mantis, on the other hand, may have to actually roll back its current fork and roll forward on a different fork just to be able to validate a checkpoint candidate. 

So I still think that it's best to stick to the model that doesn't change, which is the HotStuff paper that the `hotstuff` module is implementing, and the Advocate paper it tries to make possible, with the execution and validation methods they bring to the table. 

> Now, I realise that a consensus may be hard to reach between us on this matter. We are not a BFT federation after all

Haha, yeah, we lack the size for quorum :stuck_out_tongue:  We should ask @fishermanymc and @ranvirranaiitb, they will catch up with the thread in a few minutes no doubt :laughing: 

> before commit: applying the blocks to the state and storing the resulting state in memory
> during decide: expecting the state to be precomputed

I agree with these, I always thought that's how it would work in the `CheckpointingService` or some component used by it, and this is where the Service would have the leg up compared to the Interpreter. We can have a whole tree of tentative state in memory built up during validation (you can name it anything you like in the `checkpointing` module, as long as the `hotstuff.service.ApplicationService` can keep its naming as well), which is pruned during execution - that's why I suggested generalising the `BlockStorage` into a `KVTreeStorage`. 

You might have to recompute the state, though, if the node is restarted for example.

> This approach will probably skip BlockExecutor but will also have to validate (update ledger) during block synchronization

I think this warrants further discussions, e.g. how would you skip the `BlockExecutor`? There's also no conceptual need for validation during block sync, because the blocks have already been validated by the federation, otherwise they would not have a quroum certificate. These blocks can go directly to the execution stage when they are committed. I think this makes sense, and fighting against it would be at odds with how the BFT consensus works.

aakoshh: Sorry, don't want to sound like a broken record but found another use case to consider. 

The ChainLink whitepaper mentions HotStuff in the context of Decentralized Oracle Networks, where committee lead BFT algorithms would agree on values before putting it on the mainchain. For example the leader would propose what the stock market closing prices are for the last hour, replicas sign, then the transaction can go on the chain. This would be in contrast for example with the [Flare Network](https://blog.flare.xyz/ftso-a-breakdown/) where oracles submit values independently and the median is picked on chain (I think). 

Anyhoo, I think "validate" has a clearer meaning in this context: the replica oracle would have to consult a web server somewhere to verify whether the proposed prices reflect reality, all products exist, nothing's missing. There is no state involved since prices are independent, and the state exists outside the oracle. I think "apply block" would be more difficult to explain.

Another point I wanted to raise that in our checkpointing service, what "apply/validate" is really expected to do is send a `ValidateBlockBodyRequest` to the Interpreter. The fact that it potentially needs to apply transactions tentatively to provide the validation context is just a consequence of how validation works. On the other hand, "commit/execute" doesn't have to send this message, because at that point it doesn't matter what the Interpreter thinks, the federation has given the verdict. So even if you find you have to apply transactions on the ledger, you have to do it unconditionally.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
```diff
@@ -371,8 +422,15 @@ class ConsensusService[F[_]: Timer: Concurrent, N, A <: Agreement: Block](
 
       case CreateBlock(viewNumber, highQC) =>
         // Ask the application to create a block for us.
-        // TODO (PM-3109): Create block.
-        ???
+        appService.createBlock(highQC).flatMap {
+          case None =>
```
rtkaczyk, line 426: In what situation is `None` an acceptable response?

aakoshh: There can be a couple of reasons:
* A Prepare Q.C. means that a quorum of the federation has agreed that the BFT block pointed to by the High Q.C. is valid, confirmed by their Interpreters. But that doesn't mean that _every_ federation member had the data available to validate the block: maybe one of them lives on the "other side" of the PoW network and some transaction in the BFT block is about something they haven't received yet. If such a node becomes the next leader, they may not be able to produce a valid block. Say in the context of checkpointing, the leader is given a BFT ledger that has a last checkpoint they don't have yet, and now they have the option to nominate the next candidate. They may want to wait some time for this block to arrive, but if we time out, we can just move on to the next leader.
* It is highly unlikely, but, since this is happening in the background, the node may experience a round timeout, and jump ahead to catch up with others, clearning out earlier blocks, in which case I suppose this High Q.C. could be obsolete.
* It is possible that the `lastExecutedBlockHash` hasn't caught up with the High Q.C. yet. I'd expect that it should be possible to figure out what the state would be and produce a block, but maybe it isn't, maybe the leader just wants to pass. Although it should be possible to await the result inside the `ApplicationService` since it's the one that will be asked to do the execution as well.
* Perhaps the PoW Interpreter node crashes and doesn't respond.

rtkaczyk: Right, so it isn't a perfectly normal situation, so perhaps it should be traced? Or would you prefer to do that in `ApplicationService` where more context would be available?

aakoshh: I agree that some feedback would be good here, but you are right that the `ApplicationService` could tell much more about what the reason was. If that's acceptable, it would be more convenient for me, otherwise I'd have to change a few tracers in the robot example where a new type would be missing.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
```diff
@@ -158,49 +169,177 @@ class SyncService[F[_]: Sync, N, A <: Agreement](
...
+      _ <- syncPipe.send(SyncPipe.StatusResponse(status))
+    } yield status.viewNumber
+
+  /** Replace the state we have persisted with what we synced with the federation.
+    *
+    * Prunes old blocks, the Commit Q.C. will be the new root.
+    */
+  private def fastForwardStorage(status: Status[A], block: A#Block): F[Unit] = {
+    val blockHash = Block[A].blockHash(block)
+    assert(blockHash == status.commitQC.blockHash)
```
rtkaczyk, line 318: We'll have a lot of those. Do you think it would've been better if we had a separate QC type for each phase?
Just curious about your opinion, it's certainly not the right time to change it.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
```diff
@@ -67,6 +70,55 @@ class BlockSynchronizer[F[_]: Sync: Timer, N, A <: Agreement: Block](
...
+        alternatives: List[A#PKey]
+    ): F[Either[DownloadFailedException[A], A#Block]] = {
+      alternatives match {
+        case Nil =>
+          new DownloadFailedException(
+            quorumCertificate.blockHash,
+            sources.toVector
+          ).asLeft[A#Block].pure[F]
+
+        case source :: alternatives =>
```
rtkaczyk, line 101: Minor: I try to avoid name shadowing as much as possible. This could be named `remaining` or `nextAlternatives`

rtkaczyk (approved):
Apart from a small remark and question, can't find much to pick on. Kudos for the diagram!

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Not sure; I started with having different message types for them, but [merged them](https://github.com/input-output-hk/metronome/pull/5/commits/0004fbd40246736e6b8cf520473ddad22aa49e23) because they looked so similar.  

Maybe we could have some phantom type for it even now, something like:
```scala
type Commit = Commit.type 
...
case class QuorumCertificate[A <: Agreement, P <: VotingPhase](
    phase: P,
    viewNumber: ViewNumber,
    blockHash: A#Hash,
    signature: GroupSignature[A#PKey, (VotingPhase, ViewNumber, A#Hash), A#GSig]
)
```
The codecs can make sure that things like `Status` get the right kind then.
Not sure how we can coerce them when they appear in arbitrary places like the `Quorum` message though.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
aakoshh: Interesting, I deliberately tried to use shadowing in these loops because it happened to me that I accidentally reused the input variable which I didn't shadow and not the new one. Can't remember if a compiler plugin would have caught an unused variable in that instance. 

Why do you want to avoid it?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Tried; it's okay in the code but gets annoying in the `Gen` instances where there's a lot of `phase.next` and similar stepping.

aakoshh: Maybe it could be a tagged type, so we can use it in some places, ignore it in others.

aakoshh: Nah, that doesn't work either :-/

aakoshh: Made it work, at least up to the PR with the block executor in https://github.com/input-output-hk/metronome/pull/46

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala
rtkaczyk: 👍 I added a comment to the revelant ticket

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/sync/BlockSynchronizer.scala
rtkaczyk: Well, I do not read code line by line, I scan it. I see a symbol declaration, I see its usage, I draw the obvious conclusion. Shadows may also be tricky when moving things around, like finding that the code still compiles even though you removed a variable it used.

But that's just my opinion.

aakoshh: Good point about moving code around, I guess that's the other side of the coin I mentioned, which is accidentally using the original variable. 

Okay, renamed the variables and enabled compiler warnings for unused pattern matches and local values.

PR #38, opened 2021-04-30
PM-2941: Config parser

aakoshh:
Created a `config` module that can be used to parse HOCON config into case classes in a single step, allowing overrides from env vars. 

One problem with how we tend to use TypeSafe config is to have a method like `MyConfig.fromConfig(root.getConfig("path.to.my.config")`, which typically returns a `MyConfig` instance, or throws an error if there's a problem with the values. The other is that we can't say when this happens, as these configuration values are sometimes accessed lazily. 

The approach here does all parsing up front, so we can get all potential problems out of the way at program startup. It may be less convenient in that it's not easy to make decisions based on the values, unless some complex parser is written. A simple case of switching between identical alternatives is encoded in the `strategyDecoder`.

The way it works is that it parses the whole config into JSON, reformats the keys to be `camelCase`, so they should match those in the case classes, then uses `circe` to decode the values. See [this test](https://github.com/input-output-hk/metronome/pull/38/files#diff-e71eb41290ac00c551bf7a3dedb73acc39025ddbb71ac8ad5fa4656f391c6b7aR95) for an example.

Overrides are supported by traversing the JSON structure and looking for keys in the environment that match the path. It also supports a prefix, to make sure we're not accidentally using an env var meant for something else. For example to override `network.performance-tracing.enabled` we would create an env var called `METRONOME_NETWORK_PERFORMANCE_TRACING_ENABLED`, assuming the prefix is `METRONOME`. NB TypeSafe config also support overrides via the [CONFIG_FORCE_](https://github.com/lightbend/config#optional-system-or-env-variable-overrides) mechanism but it's a bit more involved to match the path.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -219,7 +219,11 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
       ivy"io.circe::circe-parser:${VersionOf.circe}"
     )
 
-    object test extends TestModule
+    object test extends TestModule {
+      override def ivyDeps = super.ivyDeps() ++ Agg(
+        ivy"io.circe::circe-generic:${VersionOf.circe}"
+      )
+    }
   }
 
   /** Generic HotStuff BFT library. */
--- a/build.sc
+++ b/build.sc
@@ -211,6 +211,17 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
     }
   }
 
+  /** General configuration parser, to be used by application modules. */
+  object config extends SubModule {
+    override def ivyDeps = super.ivyDeps() ++ Agg(
+      ivy"com.typesafe:config:${VersionOf.config}",
+      ivy"io.circe::circe-core:${VersionOf.circe}",
+      ivy"io.circe::circe-parser:${VersionOf.circe}"
+    )
+
+    object test extends TestModule
+  }
+
   /** Generic HotStuff BFT library. */
   object hotstuff extends SubModule {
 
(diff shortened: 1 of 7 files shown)


aakoshh:
@KonradStaniec @ntallar I'm not sure if it was one of you who lamented once about the getting configuration errors at places where they are difficult to deal with (e.g. as part of the initialisation of a lazy value), but this could be an alternative way of going about it. Something like:

```scala
def main(args: String[]): Task[ExitCode] = {
  ConfigParser.parse[NodeConfig](ConfigFactory.load().root()) match {
    case Right(Right(config)) => 
      new StdNode(config).run.use(_ => Task.never.as(ExitCode.Success))
    case other => 
      Task(log.error(other)).as(ExitCode.Error)
  }
}
```

And `StdNode` would be something like:
```
trait NodeConfigBuilder {
  def config: NodeConfig
}

class StdNode(val config: NodeConfig) extends NodeConfigBuilder { ... }
```

rtkaczyk:
Nice!

I wonder how we should deal with something like a "strategy pattern" though, e.g:

```
strategy = "foo" #one of "foo", "bar"
foo {
  ...
}
# bar { ... }
```

I guess defaults for all strategies should always be provided, and the resulting object would contain all strategies using a discriminator rather than relying on subtyping.

aakoshh:
If this strategy pattern was right at the top, we could deal with it and and optional defaults by using the Typesafe config machinery:

```scala
val strategy = conf.getString("strategy")
val startegyConf = conf.get(strategy).withFallback(conf.get("common"))
val config = ConfigParser.parse[StrategyConfig](strategyConf.root())
```

If we want this strategy to be hidden somewhere deep in config and not want to have any special handling for it, then we have other options. 

One is to list all the keys we support:
```scala
case class Strategy(
  strategy: String, 
  foo: StrategyConfig,
  bar: StrategyConfig
)
```

I see in our upstream project the `BlockchainConfig` reads all entries into a `Map[String, ConsensusConfig]` to keep it dynamic. 

As far as I see, circe `Decoder[Map[String, T]]` would also handle that bit, so we can also write a custom `Decoder[Strategy]` to turn the JSON that has a `strategy` field and other key-value entries to read it exactly like that, so:
```scala
case class Strategy(
  strategy: String, 
  strategies: Map[String, StrategyConfig]
)
```

If we don't want to surface all options in the model, then we can write a custom `Decoder[StrategyConfig]` that reads the value of the `strategy` key first, then picks the corresponding field and deserialises that using an decoder semi-auto-derived by circe for us.

To not have to repeat all default values, we can write the config like this:
```
strategy = "foo" #one of "foo", "bar"
common { ... }
foo {
  include "common"
  ... # overrides
}
# bar { ... }
```

rtkaczyk (approved):
It should be quite convenient to use :)

aakoshh:
Thank you for the question about the strategy pattern! I implemented the last option in https://github.com/input-output-hk/metronome/pull/38/commits/aa08d06401ce3932a14b91134d31f80561112560

My `include` example was wrong, that would have been the one to use if `common.conf` is in a different file, but I found another way to achieve inheritance. The tests have an example.

PR #39, opened 2021-05-05
PM-3133: Block executor

aakoshh:
Added a `BlockExecutor` component to farm out the `ExecuteBlocks` effects from the `ConsensusService`. It has an internal work queue where batches are added. A successful execution is noted down in the `ViewStateStorage`. The actual execution is delegated to the `ApplicationService`. If there's an error, we try again next time there's a new batch.

If the `ConsensusService` and the `SyncService` skip ahead to a newer Commit Quroum Certificate, and the last executed block is no longer available, the `BlockExecutor` skips blocks too. Not much else it can do. These events are traced.

### DIFF
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
@@ -40,37 +40,40 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
 
   /** Execute blocks in order, updating pesistent storage along the way. */
   private def executeBlocks: F[Unit] = {
-    def loop(lastExecutedBlockHash: A#Hash): F[Unit] =
-      executionQueue.poll
-        .flatMap {
-          case Effect.ExecuteBlocks(lastCommittedBlockHash, commitQC) =>
-            // Retrieve the blocks from the storage from the last executed
-            // to the one in the Quorum Certificate and tell the application
-            // to execute them one by one. Update the persistent view state
-            // after reach execution to remember which blocks we have truly
-            // done.
-            getBlockPath(
-              lastExecutedBlockHash,
-              lastCommittedBlockHash,
-              commitQC
-            ).flatMap {
-              case _ :: newBlockHashes =>
-                tryExecuteBatch(newBlockHashes)
-
-              case Nil =>
-                none[A#Hash].pure[F]
-            }.map {
-              _ getOrElse lastExecutedBlockHash
-            }
-        }
-        .flatMap(loop)
-
-    storeRunner
-      .runReadOnly {
-        viewStateStorage.getBundle.map(_.lastExecutedBlockHash)
-      }
-      .flatMap(loop)
+    executionQueue.poll
+      .flatMap { case Effect.ExecuteBlocks(lastCommittedBlockHash, commitQC) =>
+        // Retrieve the blocks from the storage from the last executed
+        // to the one in the Quorum Certificate and tell the application
+        // to execute them one by one. Update the persistent view state
+        // after reach execution to remember which blocks we have truly
+        // done.
+        for {
+          lastExecutedBlockHash <- getLastExecutedBlockHash
+          blockHashes <- getBlockPath(
+            lastExecutedBlockHash,
+            lastCommittedBlockHash,
+            commitQC
+          )
+          _ <- blockHashes match {
+            case _ :: newBlockHashes => tryExecuteBatch(newBlockHashes)
+            case Nil                 => ().pure[F]
+          }
+        } yield ()
+      } >> executeBlocks
   }
+
+  /** Read whatever was the last executed block that we persisted,
+    * either here or by the fast-forward synchronizer.
+    */
+  private def getLastExecutedBlockHash: F[A#Hash] =
+    storeRunner.runReadOnly {
+      viewStateStorage.getBundle.map(_.lastExecutedBlockHash)
+    }
+
+  private def setLastExecutedBlockHash(blockHash: A#Hash): F[Unit] =
+    storeRunner.runReadWrite {
+      viewStateStorage.setLastExecutedBlockHash(blockHash)
+    }
 
   /** Get the more complete path. We may not have the last executed block any more.
     *
@@ -144,13 +147,9 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
         tracers.executionSkipped(blockHash)
 
       case Some(block) =>
-        for {
-          _ <- appService.executeBlock(block)
-          _ <- storeRunner.runReadWrite {
-            viewStateStorage.setLastExecutedBlockHash(blockHash)
-          }
-          _ <- tracers.blockExecuted(blockHash)
-        } yield ()
+        appService.executeBlock(block) >>
+          setLastExecutedBlockHash(blockHash) >>
+          tracers.blockExecuted(blockHash)
     }
   }
 }
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
@@ -122,6 +122,20 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
     * The last executed block hash is used to track that it hasn't
     * been modified by the jump-ahead state sync mechanism while
     * we were executing blocks.
+    *
+    * In general we cannot expect to be able to cancel an ongoing execution,
+    * it may be in the middle of carrying out some real-world effects that
+    * don't support cancellation. To protect against race conditions between
+    * executing blocks here and the fast-forward synchroniser making changes
+    * to state, the `ApplicationService` implementation can use an exclusive
+    * lock in `executeBlock` and `syncState`.
+    *
+    * The `BlockExecutor` may not detect perfectly that it should stop executing
+    * a batch because the `lastExecutedBlockHash` changed, it might try to execute
+    * one extra block. If that is unacceptable, the `ApplicationService` can
+    * track in memory what the last executed/synced block was and check that
+    * it was indeed the parent of the next block it's asked to execute, raise
+    * an error if it's not, to stop the batch.
     */
   private def tryExecuteBatch(
       newBlockHashes: List[A#Hash],
--- a/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
+++ b/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
@@ -1,6 +1,7 @@
 package io.iohk.metronome.hotstuff.service.execution
 
 import cats.implicits._
+import cats.data.NonEmptyList
 import cats.effect.{Sync, Concurrent, ContextShift, Resource}
 import io.iohk.metronome.hotstuff.service.ApplicationService
 import io.iohk.metronome.hotstuff.service.storage.{
@@ -55,8 +56,10 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
             commitQC
           )
           _ <- blockHashes match {
-            case _ :: newBlockHashes => tryExecuteBatch(newBlockHashes)
-            case Nil                 => ().pure[F]
+            case _ :: newBlockHashes =>
+              tryExecuteBatch(newBlockHashes, commitQC)
+            case Nil =>
+              ().pure[F]
           }
         } yield ()
       } >> executeBlocks
@@ -107,7 +110,8 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
     * Return the last successfully executed hash, if any.
     */
   private def tryExecuteBatch(
-      newBlockHashes: List[A#Hash]
+      newBlockHashes: List[A#Hash],
+      commitQC: QuorumCertificate[A]
   ): F[Option[A#Hash]] = {
     def loop(
         newBlockHashes: List[A#Hash],
@@ -118,7 +122,11 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
           lastExecutedBlockHash.pure[F]
 
         case blockHash :: newBlockHashes =>
-          executeBlock(blockHash).attempt.flatMap {
+          executeBlock(
+            blockHash,
+            commitQC,
+            NonEmptyList(blockHash, newBlockHashes)
+          ).attempt.flatMap {
             case Left(ex) =>
               // If a block fails, return what we managed to do so far,
               // so we can re-attempt it next time if the block is still
@@ -139,7 +147,14 @@ class BlockExecutor[F[_]: Sync, N, A <: Agreement](
     * Be prepared that it may not exist, if execution took so long that
     * the `SyncService` skipped ahead to the latest Commit Q.C.
     */
-  private def executeBlock(blockHash: A#Hash): F[Unit] = {
+  private def executeBlock(
+      blockHash: A#Hash,
+      commitQC: QuorumCertificate[A],
+      commitPath: NonEmptyList[A#Hash]
+  ): F[Unit] = {
+    assert(commitPath.head == blockHash)
+    assert(commitPath.last == commitQC.blockHash)
+
     storeRunner.runReadOnly {
       blockStorage.get(blockHash)
     } flatMap {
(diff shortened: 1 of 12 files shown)


metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/storage/BlockStorageProps.scala
```diff
@@ -238,6 +254,53 @@ object BlockStorageProps extends Properties("BlockStorage") {
       data.store.getPathFromRoot(nonExisting.id).isEmpty
   }
 
+  property("getPathFromAncestor") = forAll(
+    for {
+      prefix <- genNonEmptyBlockTree
+      ancestor = prefix.last
+      postfix    <- genNonEmptyBlockTree(ancestor.id)
+      descendant <- Gen.oneOf(postfix)
+      data = TestData(prefix ++ postfix)
```
rtkaczyk, line 263: Do we need the ancestors of the `ancestor`? Why not just generate a single block tree?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
```diff
@@ -0,0 +1,287 @@
...
+          }
+
+        } yield {
+          "executes from the root" |: executedBlockHashes == pathFromRoot.tail
+        }
+      }
+    }
+  }
+
+  property("executeBlocks - from last") = forAll { (fixture: TestFixture) =>
```
rtkaczyk, line 210: Should we set `maxBatches = 1` since we're only enqueuing the last one?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  property("executeBlocks - from pruned") = forAll { (fixture: TestFixture) =>
+    run {
+      fixture.resources.use { case (blockSychronizer, _) =>
+        val lastBatch             = fixture.batches.last
+        val lastExecutedBlockHash = lastBatch.lastExecutedBlockHash
+        for {
+          _ <- fixture.storeRunner.runReadWrite {
+            TestBlockStorage.pruneNonDescendants(lastExecutedBlockHash)
```
rtkaczyk, line 246: I don't quite understand this test. Do we even populate the storage with such non-descendants?

rtkaczyk:
I feel we've discussed the main code sufficiently when discussing PM-3108/3110, and I'm happy with it 👍 , so I'm focusing my review on the tests

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/storage/BlockStorageProps.scala
aakoshh: I wanted to make sure that I can pick any 2 blocks in the tree which are related, and that the traversal will only go between them, not higher to the root for example. 

Conceptually I only need a single block tree, but the generator only returns a list of blocks, it's not obvious how to pick an ancestor and a descendant from that, unless one of them is the root. By doing it in 2 stages, I know that every block in the `postfix` is a descendant of the `ancestor`, so I can pick any of them, and `ancestor` may or may not be the `root` of the `prefix`. 

I figured it's easier to generate the data in the most convenient way, so there doesn't need to be any logic in the test.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
aakoshh: Sorry, there aren't many comments in the setup.

Yeah, we generate a [prefixTree](https://github.com/input-output-hk/metronome/pull/39/files/7fd73c729ad087906359aa78ad637b9ce3b41108#diff-43a00c2f1553bcab8cd7e482b1c3d5861d3cc10c91528c904910a85a46e909f4R176) first and then between 1 and 5 extension trees; this test picks `lastBatch.lastExecutedBlockHash` as the new root, so after pruning we shall only have the last extension tree. The non-descendants are the ancestors, uncles and cousins in the prefix tree and any previous extensions.

The test checks that even if we prune the block that were the last executed ones according to our view state, the `BlockExecutor` can recover by going from whatever the last committed block before the current batch was.

This is implemented [here](https://github.com/input-output-hk/metronome/pull/39/files/7fd73c729ad087906359aa78ad637b9ce3b41108#diff-6c7370a74c3203456a648dd0b2eb746f549b452dd00006152771767883fba0dbR111-R114) but I can't remember why I thought a situation like this would arise. I only implemented pruning for the robot and it prunes blocks [relative to the last executed hash](https://github.com/input-output-hk/metronome/pull/42/files#diff-10b56ac2536b604ca590b5fe6859da38983d0023e5199eb3d5659d1d9cee12ddR419-R421), so this shouldn't happen. Maybe it was before I added the reading of the last hash from the view state before each batch :thinking:

aakoshh: Yeah, that's how it [started](https://github.com/input-output-hk/metronome/pull/39/commits/50493682ce2decf005352e772bdb178c8529aabb#diff-6c7370a74c3203456a648dd0b2eb746f549b452dd00006152771767883fba0dbR43-R56), and I think I was just being defensive, thinking what would happen if someone would prune away the blocks we thought we last executed, surely they would keep at least the last committed one.

aakoshh: plans within plans

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/storage/BlockStorageProps.scala
rtkaczyk: > that the traversal will only go between them, not higher to the root for example.

Makes sense, but we're not really testing that, are we? I mean we're only looking at the result, not at how it's produced. In fact, maybe it could actually test the traversal if the ancestor was always the root, as it would produce an error if we try to access its parent?

Not a big deal, just pointing out

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
rtkaczyk: Sorry, but I fail to see how the other batches are relevant here:

> at the time it takes on the next batch

```
_ <- blockSychronizer.enqueue(lastBatch)
```

Since we're executing a single batch, there is no next batch. Am I missing something?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
rtkaczyk: I'm still not quite clear. If I understand correctly the initial store is a chain of blocks, so what is there to be pruned?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/storage/BlockStorageProps.scala
aakoshh: For a few minutes I didn't know what you meant by not testing it. Now I understand you meant that it _could_ have accessed the parent during traversal, just not returned it. 

Sorry, I didn't mean to check what the traversal does behind the scenes, only that we return the `ancestor` as the `head` of `path` (which we [check](https://github.com/input-output-hk/metronome/pull/39/files/7fd73c729ad087906359aa78ad637b9ce3b41108#diff-040272a08fe005f889fce95f80d70b6c857e6cf9a776c14a0fd001c98628a9fcR274-R275)), not just by accident because it doesn't have a parent to go to, but because that's what the result is supposed to be. 

The `ancestor` _can_ be the root of the prefix (if it's a single element tree), so if it threw an exception trying to access its parent it should fail the test in some cases. I assume you didn't want to suggest by "always the root" that this should be the only way we should test it?

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
aakoshh: No, you're not missing anything, it's only using the last batch, the others may only be there to emphasise that their presence should make no difference. 

Sorry if I didn't make myself clear: I agree with you that setting `maxBatches = 1` would achieve the same effect. I just didn't want to specify something that doesn't have any bearing on whether the test should work, to avoid making it look like that's a crucial setting (like it is when I set `minBatches = 2` in the failure case). The only effect should be that test generation may become quicker.

Now, if there wasn't the `prefixTree` in the data generation (a 0th batch if you will), then it could have an effect, because in that case `maxBatches = 1` would mean always going from the root.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
aakoshh: The initial store has all the block tree that was generated, and the batches are what will tell what the committed chain from that is. 

Say it has a tree like this were `0*` is the `prefixTree`, `1*` is the first and `2*` is the second extension:
```
                            15
                          /
01 - 02 - 04 - 11 - 12 - 14 - 16 
   \                 \ 
    03                13 - 17 - 21 - 22 - 23
                                      \
                                       24 - 25
```
Then the fixture would have 2 batches: `(04, 17)` and `(17, 25)`. In each pair the 1st member is what has been committed in the previous batch, the 2nd is the new committed block. The pruning in this case would get rid of anything before `17`. 

> the call to pruneNonDescendants is kind of arbitrary. What if it's called with a different blockHash than what we expect?

You are right, the pruning here is specific to one of the conditions that the `BlockExecutor` knows how to handle. If we pruned from `13`, it wouldn't execute anything. The batches mean that we only have commits for `04`, `17` and `25`. The recovery mechanism in the executor works by assuming that even if something went wrong during the execution, the federation will keep working and give us a new commit at some point, at which point we can again try and see if we can find that block. 

In our example, the `lastExecutedBlockHash` would be `01` but for some reason someone pruned to `13`, so the batch of `(04, 17)` would fail because neither `01` nor `04` exist. But then would come batch `(17, 25)` and we would see that `17` indeed exists in storage. We ask the `ApplicationService` to execute `21`, `22`, `24` and `25`, because it always assumes that the first item of the batch has been done before. If `21` fails, then we are stuck, but if we indeed synced to `17` somehow, then it can resume from there  and update `lastExecutedBlockHash`.

The failure scenario is tested further down. This test only checks that if the `lastExecutedHash` in `ViewStateStorage` doesn't exist but the one in the batch does, we use that one to attempt to execute from. 

> maybe we should add a test for the executionSkipped case?

Yeah, good idea, will add one :+1:

aakoshh: Added the test for `executionSkipped` in https://github.com/input-output-hk/metronome/pull/39/commits/40d1a00a78120627012c88cee222526062bb5a8b

Thank you for raising it, it caught a bug as well, the first block that failed the compare-and-set wasn't traced.

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/storage/BlockStorageProps.scala
rtkaczyk: Yeah, makes sense 👍

metronome/hotstuff/service/test/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutorProps.scala
rtkaczyk: > test generation may become quicker.

That's what I was after

> avoid making it look like that's a crucial setting

but I'm fine with this reasoning

aakoshh:
@rtkaczyk on the point of skipped executions I got into a dilemma about this scenario:
1. The block executor starts executing a batch `A`, `B`, `C` and `D`.
2. Things are going slow, currently executing `B`.
3. Meanwhile, the consensus gets out of sync and decides it needs to jump ahead.
4. The federation is on block `F`, we sync there by downloading the state of `F`.
5. Race condition happens: `B` is finished and set as the last executed block hash, then `F` is set as the last executed block hash, but only after the executor set `B`, so the executor carries on with the next block.
6. The executor executes `C`, tries to set the last hash to `C` but realises it's not `B` any more, so it skips `D`. 

Executing `C` without realising that we are already in state `F` could be bad. My first thought was that this can be handled by the `ApplicationService`: 
* It can use a `Semaphore` to make sure it's not doing `stateSync` and `executeBlock` at the same time. However that doesn't solve the above issue alone.
* It can maintain a `lastExecutedBlockHashRef` of its own and check in `executeBlock` that `block.parentHash == lastExecutedBlockHash`, and also set it in `stateSync`. 

I thought this is too much lore to get right, so I moved the call to `stateSync` and the subsequent `fastForwardStorage` from `SyncService` to the `BlockExecutor`, and added a semaphore around the critical sections. This should make sure that we would not be executing `C` because first we execute `A`,`B`,`C` and `D` completely before syncing to `E`.

I know you said the `BlockExecutor` does too much, and now it does even more, but I reckon the overall semantics are easier to follow this way.

I also modified `ApplicationService.syncState` to return a flag to indicate whether it's okay to prune away old blocks, although this could be a constructor argument as well.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/execution/BlockExecutor.scala
```diff
@@ -0,0 +1,304 @@
...
+    *
+    * Prunes old blocks, the Commit Q.C. will be the new root.
+    */
+  private def fastForwardStorage(
+      block: A#Block,
+      canPrune: Boolean
+  ): F[Unit] = {
+    val blockHash = Block[A].blockHash(block)
+
+    val prune = for {
```
rtkaczyk, line 255: 1. Why are we not using `blockStorage.pruneNonDescendants` here? Is it because there may be no link between the root and the `blockhash`? 
2. What about the safe deletion order?
3. Should we maybe put this into `BlockStorage` as an alternative pruning method so that the difference in use cases can be more easily spotted?

aakoshh: That got me scratching my head, I should add a comment. 

1. You're exactly right, when `syncState` is invoked we can expect that there will be a gap between what we have in the block storage and the block to which we just synced, so there is (probably) no way to traverse to the root from the new block. That's why I added the root block hash explicitly to the view state, so we can go from the other direction. The `skip` is used purely as a safeguard.
2. Arguably at this point we don't care about the deletion order because we assume we are dealing with a forest, and we just want to purge the old tree and keep the new (single) tree. It will be executed as one batch against RocksDB. 
3. The thing is that the block store doesn't store the root, we rely on the view store here to figure out where to go from. We discussed earlier that the block store _can_ store a forest, and it's up to the user to keep track of pointers into each tree. The view store knows a bit more, we know we only care about this one tree. We could move a method that just purges all descendants of a certain block, but there I would try to be more careful about the deletion order as I could not assume the block is the root... maybe a method that goes up to the root and kills the root, a `purgeTree`. Should we do that?

aakoshh: Added in https://github.com/input-output-hk/metronome/pull/39/commits/0b8b29d90c4c45ee2556482664880d0bc0d210cf
Thanks for the suggestion!

rtkaczyk: I like it better that way :+1:

rtkaczyk:
It was on my list to look for race conditions regarding `lastExecutedBlockhash`, but it seems you fixed it before I got a change. Cool 🎉

rtkaczyk (approved):
Just so you know, I've made my peace with the BlockExecutor being what it is. It's certainly not easy to grasp, but should make implementing applications easier 💯

PR #40, opened 2021-05-05
[PM-3064,3065] signing

rtkaczyk:
Secp256k1-based group and partial signing

### DIFF
--- a/metronome/crypto/src/io/iohk/metronome/crypto/Secp256k1Utils.scala
+++ b/metronome/crypto/src/io/iohk/metronome/crypto/Secp256k1Utils.scala
@@ -1,28 +1,14 @@
 package io.iohk.metronome.crypto
 
 import java.security.SecureRandom
-import org.bouncycastle.crypto.AsymmetricCipherKeyPair
-import org.bouncycastle.crypto.params.ECPublicKeyParameters
-import scodec.bits.BitVector
 
 object Secp256k1Utils {
 
   def generateKeyPair(
       secureRandom: SecureRandom
-  ): AsymmetricCipherKeyPair = {
-    io.iohk.ethereum.crypto.generateKeyPair(secureRandom)
-  }
-
-  /** Returns secp256k1 public key bytes in uncompressed form, with compression indicator stripped
-    */
-  def keyPairToUncompressed(keyPair: AsymmetricCipherKeyPair): BitVector = {
-    BitVector(
-      keyPair.getPublic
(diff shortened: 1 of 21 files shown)


metronome/crypto/src/io/iohk/metronome/crypto/Secp256k1Utils.scala
```diff
@@ -1,28 +1,14 @@
 package io.iohk.metronome.crypto
 
 import java.security.SecureRandom
-import org.bouncycastle.crypto.AsymmetricCipherKeyPair
-import org.bouncycastle.crypto.params.ECPublicKeyParameters
-import scodec.bits.BitVector
 
 object Secp256k1Utils {
```
rtkaczyk, line 5: Will remove once we choose the version

metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/basic/Secp256k1Signing.scala
```diff
@@ -0,0 +1,109 @@
...
+    val areUniqueSigners     = signers.size == signature.sig.size
+    val areFederationMembers = (signers -- federation.publicKeys).isEmpty
+    val isQuorumReached      = signers.size >= federation.quorumSize
+
+    areUniqueSigners && areFederationMembers && isQuorumReached
+  }
+
+  // For improved readability these provide implicit conversions
+  // between the generic Agreement types and the concrete implementation
+  // types based on the implicit type equality evidence
```
aakoshh, line 95: I see, nice trick! 

Well, I think the `Secp256k1Agreement` approach has the benefit of making it just slightly easier to create a new type of agreement, by just mixing in the right trait, without having to understand which exact types are used. It also means the author of `Sepc26k1Agreement` can change some thing around without affecting all the `Agreement` subtypes in existence that would otherwise get compilation errors with the missing witnesses. 

Maybe this technique is best used in moderation, and we should use the more pragmatic approach in the other PR. 

WDYT?

rtkaczyk: Yes, I agree. Closing this

aakoshh: We can always add them back and keep the trait as well, delegating to the method :stuck_out_tongue:

PR #42, opened 2021-05-11
PM-3241: Example application - Robot movement

aakoshh:
Created an `examples` module to demonstrate HotStuff. The demo use case is moving around a robot by consensus. 

The PR is rather big, there are some bug fixes in it but mostly boilerplate. I at least managed to generalise the codecs (not using RLP here though), so a similar program could be smaller. Probably there are more things we can generalise in application composition as well.

The instructions for how to run it are in the [README](https://github.com/input-output-hk/metronome/tree/pm-3241-example-app#robot).

I wanted to have a [nice display](https://eed3si9n.com/console-games-in-scala) but the PR is already huge and the console is pared down to the level where it's easy to see when there is or isn't consensus:

image


rtkaczyk (approved):
I have reviewed this to a limited extent, focusing on:
- `RobotCompositon`
- changes in `ConsensusService`
- running the app 

 I'm confident this is a great demonstration on how to use the underlying services

PR #43, opened 2021-05-19
PM-3241: Fix RocksDB locking to be synchronous.

aakoshh:
Discovered as part of testing https://github.com/input-output-hk/metronome/pull/42

I should not have missed this: when using `Task`, multiple operations are executed synchronously in `flatMap` (unlike with `Future` which does everything asynchronously) until the thread yields and processes other tasks. The way locking and unlocking was done was susceptible to the lock being taken on one thread and the unlock attempted on another, resulting in rather elusive deadlocks. In the end I saw `RocksDBStore` mentioned in thread dumps, with threads waiting on read and write locks, and it turned out sometimes it cannot unlock. 

Changed the database to use the `Eval` monad in compiling the `Free` operations, and only at the edge turn them into `F[A]`.

### DIFF
--- a/metronome/rocksdb/src/io/iohk/metronome/rocksdb/RocksDBStore.scala
+++ b/metronome/rocksdb/src/io/iohk/metronome/rocksdb/RocksDBStore.scala
@@ -3,8 +3,7 @@ package io.iohk.metronome.rocksdb
 import cats._
 import cats.implicits._
 import cats.data.ReaderT
-import cats.effect.{Resource, Sync}
-import cats.free.Free.liftF
+import cats.effect.{Resource, Sync, ContextShift, Blocker}
 import io.iohk.metronome.storage.{
   KVStore,
   KVStoreOp,
@@ -44,9 +43,10 @@ import scala.annotation.nowarn
   * locking could be performed in their respective middle-layers, before they forward the
   * query for execution to this class.
   */
-class RocksDBStore[F[_]: Sync](
-    db: RocksDBStore.DBSupport[F],
-    lock: RocksDBStore.LockSupport[F],
+class RocksDBStore[F[_]: Sync: ContextShift](
+    db: RocksDBStore.DBSupport,
+    lock: RocksDBStore.LockSupport,
+    blocker: Blocker,
     handles: Map[RocksDBStore.Namespace, ColumnFamilyHandle]
 ) {
 
@@ -62,7 +62,7 @@ class RocksDBStore[F[_]: Sync](
   // Type aliases to support the `~>` transformation with types that
   // only have 1 generic type argument `A`.
   type Batch[A] =
-    ({ type L[A] = ReaderT[F, BatchEnv, A] })#L[A]
+    ({ type L[A] = ReaderT[Eval, BatchEnv, A] })#L[A]
 
   type KVNamespacedOp[A] =
     ({ type L[A] = KVStoreOp[Namespace, A] })#L[A]
@@ -71,25 +71,23 @@ class RocksDBStore[F[_]: Sync](
     ({ type L[A] = KVStoreReadOp[Namespace, A] })#L[A]
 
   /** Execute the accumulated write operations in a batch. */
-  private val writeBatch: ReaderT[F, BatchEnv, Unit] =
+  private val writeBatch: ReaderT[Eval, BatchEnv, Unit] =
     ReaderT { batch =>
       if (batch.hasPut() || batch.hasDelete())
         db.write(batch) >>
-          Sync[F].delay {
-            batch.clear()
-          }
+          Eval.always(batch.clear())
       else
-        ().pure[F]
+        ().pure[Eval]
     }
 
   /** Execute one `Get` operation. */
-  private def read[K, V](op: Get[Namespace, K, V]): F[Option[V]] = {
+  private def get[K, V](op: Get[Namespace, K, V]): Eval[Option[V]] = {
     for {
       kbs  <- encode(op.key)(op.keyEncoder)
-      mvbs <- db.read(handles(op.namespace), kbs)
+      mvbs <- db.get(handles(op.namespace), kbs)
       mv <- mvbs match {
         case None =>
-          none.pure[F]
+          none.pure[Eval]
 
         case Some(bytes) =>
           decode(bytes)(op.valueDecoder).map(_.some)
@@ -118,7 +116,7 @@ class RocksDBStore[F[_]: Sync](
 
           case op @ Get(_, _) =>
             // Execute any pending deletes and puts before performing the read.
-            writeBatch >> ReaderT.liftF(read(op))
+            writeBatch >> ReaderT.liftF(get(op))
 
           case op @ Delete(n, k) =>
             ReaderT { batch =>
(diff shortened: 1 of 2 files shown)


metronome/rocksdb/src/io/iohk/metronome/rocksdb/RocksDBStore.scala
```diff
@@ -425,31 +465,55 @@ object RocksDBStore {
...
-      }(_ => fa) { _ =>
+    def withLockUpgrade[A](fa: Eval[A]): Eval[A] =
+      bracket(
+        unlockRead >> lockWrite,
         lockRead >> unlockWrite
+      )(fa)
+
+    private def bracket[A](lock: Eval[Unit], unlock: Eval[Unit])(
+        evalA: Eval[A]
+    ): Eval[A] = Eval.always {
```
KonradStaniec, line 476: the eval docs mention:
```
It is also not good style to create Eval instances whose computation involves calling .value on another Eval instance -- this can defeat the trampolining and lead to stack overflows.
```

do you think it can be a problem here ?

metronome/rocksdb/src/io/iohk/metronome/rocksdb/RocksDBStore.scala
```diff
@@ -425,31 +465,55 @@ object RocksDBStore {
      * See here for the rules up (non-)upgrading and downgrading:
      * https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html
      */
-    def withLockUpgrade[A](fa: F[A]): F[A] =
-      Sync[F].bracket {
-        unlockRead >> lockWrite
-      }(_ => fa) { _ =>
+    def withLockUpgrade[A](fa: Eval[A]): Eval[A] =
```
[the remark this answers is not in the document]
aakoshh: Oops, thanks man! Looks like I forgot to add it back to the `nonBatchingCompiler`, but it should be used, to protect the consistency of larger read-write operations from these sporadic writes. Added it back.

metronome/rocksdb/src/io/iohk/metronome/rocksdb/RocksDBStore.scala
aakoshh: Thanks for pointing out this out, I didn't see it in the docs. I hope it wouldn't cause a stack overflow because there shouldn't be any recursion, the computations in `evalA` don't take out further locks (except for the upgrade). I couldn't find another way to add `finally`, `Eval` is not an instance of `MonadError`, I guess because it cannot suspend an error.

KonradStaniec: tbh I also hope that it will not  :) I am wondering maybe monix `Coeval` would be the answer here ? From what I remember it has `.doOnFinish` method (and probably more)

aakoshh: That's a good shout, it even [says in the docs](https://monix.io/docs/current/eval/coeval.html#comparison-with-cats-eval) that while it's inspired by `Eval` it does error handling.

aakoshh: I wanted to stick with `cats`, not sure if it's worth bringing in monix into the picture just for this :thinking:

aakoshh: Apparently it [used to have](https://github.com/typelevel/cats/issues/1639) `MonadError` but was removed.

KonradStaniec: it got me thinking, as `Eval` does not implement `MonedError` (or in general is not designed to work with operations which can throw errors) does that mean that possible error thrown by db operation will not be propagated in `F` context or the fact that we either way wrap everyting in `Sync[F].delay(evalA.value) kinda cover this case ?

aakoshh: Wrapping it in `Sync` should cover it. It can suspend errors like this:
```
scala> Eval.always(sys.error("Boom!"))
val res0: cats.Eval[Nothing] = cats.Always@1cc93da4

scala> res0.value
java.lang.RuntimeException: Boom!
  at scala.sys.package$.error(package.scala:27)
  at $anonfun$res0$1(<console>:1)
  at cats.Always.value(Eval.scala:171)
  ... 32 elided
scala> Eval.defer(sys.error("Boom!"))
val res3: cats.Eval[Nothing] = cats.Eval$$anon$5@914ab92

scala> res3.value
java.lang.RuntimeException: Boom!
  at scala.sys.package$.error(package.scala:27)
  at $anonfun$res3$1(<console>:1)
  at cats.Eval$.loop$1(Eval.scala:341)
  at cats.Eval$.cats$Eval$$evaluate(Eval.scala:363)
  at cats.Eval$Defer.value(Eval.scala:262)
  ... 32 elided
```

I'm not sure why it doesn't have a `MonadError`, perhaps that would require that it should be able to represent an error as a pure value that can be inspected for what it is without having to throw and catch the error.

KonradStaniec: sure, so as for me, if you are sure that we not risking stackoverflows here, then pr looks good to me  :+1:

aakoshh: I did this experiment:
```console
scala> import cats.Eval
import cats.Eval

scala> val e = Eval.always(1)
val e: cats.Eval[Int] = cats.Always@d1f23a0

scala> import cats.implicits._
import cats.implicits._

scala> val es = List.fill(1000000)(e)
val es: List[cats.Eval[Int]] = List(cats.Always@d1f23a0, cats.Always@d1f23a0, cats.Always@d1f23a0, ...

scala> val esf1 = es.foldLeft(Eval.always(0)) { case (ea, eb) => ea.flatMap(a => eb.map(b => a+b)) }
val esf1: cats.Eval[Int] = cats.Eval$$anon$1@2f8cb4c9

scala> esf1.value
val res3: Int = 1000000

scala> val esf2 = es.foldLeft(Eval.always(0)) { case (ea, eb) => Eval.always(ea.value + eb.value) }
val esf2: cats.Eval[Int] = cats.Always@67710b92

scala> esf2.value
java.lang.StackOverflowError
  at cats.Always.value(Eval.scala:171)
  at $anonfun$esf$3(<console>:1)
  at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.scala:17)
  at cats.Always.value(Eval.scala:171)
  at $anonfun$esf$3(<console>:1)
 ...

scala> Eval.always(esf1.value)
val res6: cats.Eval[Int] = cats.Always@44640a84

scala> res6.value
val res7: Int = 1000000
```

So it looks like stacking `Eval.always(someEval.value)` can indeed cause a stack overflow, but just entering a top level one like the one used in `bracket` doesn't defeat the trampoline.

PR #45, opened 2021-05-31
[PM-3108] Update Ledger

rtkaczyk:
The motivation is that since in the normal flow a block is first validated and the resulting state is saved in memory then it should not be necessary for `BlockExecutor` to prepare a batch of blocks to execute. The chosen design however should work with or without it. Having the intermediate Ledgers saved in memory should not require any additional queries to the persistent storage during execution, as long as the service run uninterrupted.

The in memory map of Ledgers should always hold the last committed state, being reset to that single value after each commit. The associated block headers can be used to construct the checkpoint certificate.

### DIFF
--- a/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Ledger.scala
+++ b/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Ledger.scala
@@ -1,7 +1,5 @@
 package io.iohk.metronome.checkpointing.models
 
-import io.iohk.metronome.core.Validated
-
 /** Current state of the ledger after applying all previous blocks.
   *
   * Basically it's the last checkpoint, plus any accumulated proposer blocks
@@ -21,8 +19,8 @@ case class Ledger(
     * by this point, so we know for example that the new checkpoint is
     * a valid extension of the previous one.
     */
-  def update(transaction: Validated[Transaction]): Ledger =
-    (transaction: Transaction) match {
+  def update(transaction: Transaction): Ledger =
+    transaction match {
       case t @ Transaction.ProposerBlock(_) =>
         if (proposerBlocks.contains(t))
           this
@@ -32,6 +30,9 @@ case class Ledger(
       case t @ Transaction.CheckpointCandidate(_) =>
         Ledger(Some(t), Vector.empty)
     }
+
+  def update(transactions: Iterable[Transaction]): Ledger =
+    transactions.foldLeft(this)(_ update _)
 }
 
 object Ledger extends RLPHashCompanion[Ledger]()(RLPCodecs.rlpLedger) {
--- a/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
+++ b/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
@@ -1,0 +1,239 @@
+package io.iohk.metronome.checkpointing.service
+
+import cats.data.{NonEmptyList, NonEmptyVector, OptionT}
+import cats.effect.concurrent.Ref
+import cats.effect.{Concurrent, Resource, Sync}
+import cats.implicits._
+import io.iohk.metronome.checkpointing.CheckpointingAgreement
+import io.iohk.metronome.checkpointing.models.{
+  Block,
+  CheckpointCertificate,
+  Ledger
+}
+import io.iohk.metronome.checkpointing.service.CheckpointingService.LedgerTree
+import io.iohk.metronome.checkpointing.service.storage.LedgerStorage
+import io.iohk.metronome.crypto.ECPublicKey
+import io.iohk.metronome.hotstuff.consensus.basic.{Phase, QuorumCertificate}
+import io.iohk.metronome.hotstuff.service.ApplicationService
+import io.iohk.metronome.hotstuff.service.storage.{
+  BlockStorage,
+  ViewStateStorage
+}
+import io.iohk.metronome.storage.KVStoreRunner
+
+import scala.annotation.tailrec
+
+class CheckpointingService[F[_]: Sync, N](
+    ledgerTree: Ref[F, LedgerTree],
+    lastExecutedHeader: Ref[F, Block.Header],
+    ledgerStorage: LedgerStorage[N],
+    blockStorage: BlockStorage[N, CheckpointingAgreement]
+)(implicit storeRunner: KVStoreRunner[F, N])
+    extends ApplicationService[F, CheckpointingAgreement] {
+
+  override def createBlock(
+      highQC: QuorumCertificate[CheckpointingAgreement]
+  ): F[Option[Block]] = ???
+
+  override def validateBlock(block: Block): F[Option[Boolean]] = {
+    val ledgers = for {
+      nextLedger <- OptionT(projectLedger(block))
+      tree       <- OptionT.liftF(ledgerTree.get)
+      prevLedger <- tree.get(block.header.parentHash).map(_._1).toOptionT[F]
+    } yield (prevLedger, nextLedger)
+
+    ledgers.value.flatMap {
+      case Some((prevLedger, nextLedger))
+          if nextLedger.hash == block.header.postStateHash =>
+        validateTransactions(block.body, prevLedger)
+
+      case _ => false.some.pure[F]
+    }
+  }
+
+  private def validateTransactions(
+      body: Block.Body,
+      ledger: Ledger
+  ): F[Option[Boolean]] = {
+    //TODO: Validate transactions PM-3131/3132
+    true.some.pure[F]
+  }
+
+  override def executeBlock(
+      block: Block,
+      commitQC: QuorumCertificate[CheckpointingAgreement],
+      commitPath: NonEmptyList[Block.Hash]
+  ): F[Boolean] = {
+    require(commitQC.phase == Phase.Commit, "Commit QC required")
+    projectLedger(block).flatMap {
+      case Some(ledger) =>
+        if (block.hash != commitQC.blockHash)
+          saveLedger(block.header, ledger).as(true)
+        else
+          for {
+            //TODO: PM-3110:
+            // chkpOpt <- constructCheckpoint(ledger, commitQC)
+            _ <- saveLedger(block.header, ledger)
+            //TODO: PM-3110:
+            // _ <- chkpOpt.map(pushCheckpoint).getOrElse(().pure[F])
+          } yield true
+
+      case None =>
+        Sync[F].raiseError(
+          new IllegalStateException(s"Could not execute block: ${block.hash}")
+        )
+    }
+  }
+
+  private def projectLedger(block: Block): F[Option[Ledger]] = {
+    (for {
+      ledgers    <- ledgerTree.get
+      execHeight <- lastExecutedHeader.get.map(_.height)
+    } yield {
+      def loop(block: Block): OptionT[F, Ledger] = {
+        def doUpdate(ledger: Ledger) =
+          OptionT.liftF(updateLedgerByBlock(ledger, block))
+
+        ledgers.get(block.header.parentHash) match {
+          case Some((oldLedger, _)) =>
+            doUpdate(oldLedger)
+
+          case None if block.header.height <= execHeight =>
+            OptionT.none
+
+          case None =>
+            for {
+              parent    <- OptionT(getBlock(block.header.parentHash))
+              oldLedger <- loop(parent)
+              newLedger <- doUpdate(oldLedger)
+            } yield newLedger
+        }
+      }
+
+      ledgers
+        .get(block.hash)
+        .map(_._1)
+        .toOptionT[F]
+        .orElse(loop(block))
+        .value
+    }).flatten
+  }
+
+  private def updateLedgerByBlock(
+      oldLedger: Ledger,
+      block: Block
+  ): F[Ledger] = {
+    val newLedger = oldLedger.update(block.body.transactions)
+    ledgerTree
+      .update { tree =>
+        if (tree.contains(block.header.parentHash))
+          tree + (block.hash -> (newLedger, block.header))
+        else
+          tree
+      }
+      .as(newLedger)
+  }
+
+  private def getBlock(hash: Block.Hash): F[Option[Block]] =
+    storeRunner.runReadOnly(blockStorage.get(hash))
+
+  private def saveLedger(header: Block.Header, ledger: Ledger): F[Unit] = {
+    storeRunner.runReadWrite {
+      ledgerStorage.put(ledger)
+    } >>
+      ledgerTree.update(clearLedgerTree(header, ledger)) >>
+      lastExecutedHeader.set(header)
+  }
+
+  /** Makes the `commitHeader` and the associated 'ledger' the root of the tree,
+    * while retaining any descendants of the `commitHeader`
+    */
+  private def clearLedgerTree(commitHeader: Block.Header, ledger: Ledger)(
+      ledgerTree: LedgerTree
+  ): LedgerTree = {
+
+    @tailrec
+    def loop(
+        oldTree: LedgerTree,
+        newTree: LedgerTree,
+        height: Long
+    ): LedgerTree =
+      if (oldTree.isEmpty) newTree
+      else {
+        val (higherLevels, currentLevel) = oldTree.partition {
+          case (_, (_, hd)) => hd.height > height
+        }
+        val children = currentLevel.filter { case (_, (_, hd)) =>
+          newTree.contains(hd.parentHash)
+        }
+        loop(higherLevels, newTree ++ children, height + 1)
+      }
+
+    loop(
+      ledgerTree.filter { case (_, (_, hd)) =>
+        hd.height > commitHeader.height
+      },
+      Map(commitHeader.hash -> (ledger, commitHeader)),
+      commitHeader.height + 1
+    )
+  }
+
+  private def constructCheckpoint(
+      ledger: Ledger,
+      commitQC: QuorumCertificate[CheckpointingAgreement]
+  ): F[Option[CheckpointCertificate]] =
+    ??? //TODO: PM-3110
+
+  private def pushCheckpoint(checkpoint: CheckpointCertificate): F[Unit] =
+    ??? //TODO: PM-3137
+
+  override def syncState(
+      sources: NonEmptyVector[ECPublicKey],
+      block: Block
+  ): F[Boolean] = ???
+}
+
+object CheckpointingService {
+  type LedgerTree = Map[Block.Hash, (Ledger, Block.Header)]
+
+  def apply[F[_]: Concurrent, N](
+      ledgerStorage: LedgerStorage[N],
+      blockStorage: BlockStorage[N, CheckpointingAgreement],
+      viewStateStorage: ViewStateStorage[N, CheckpointingAgreement]
+  )(implicit
+      storeRunner: KVStoreRunner[F, N]
+  ): Resource[F, CheckpointingService[F, N]] = {
+    val lastExecuted: F[(Block, Ledger)] =
+      storeRunner.runReadOnly {
+        val query = for {
+          blockHash <- OptionT.liftF(
+            viewStateStorage.getLastExecutedBlockHash
+          )
+          block <- OptionT(blockStorage.get(blockHash))
+          //a genesis (empty) state should be present in LedgerStorage on first run
+          ledger <- OptionT(ledgerStorage.get(block.header.postStateHash))
+        } yield (block, ledger)
+        query.value
+      } >>= {
+        _.toOptionT[F].getOrElseF {
+          Sync[F].raiseError(
+            new IllegalStateException("Last executed block/state not found")
+          )
+        }
+      }
+
+    val service = for {
+      (block, ledger) <- lastExecuted
+      ledgerTree      <- Ref.of(Map(block.hash -> (ledger, block.header)))
+      lastExec        <- Ref.of(block.header)
+    } yield new CheckpointingService[F, N](
+      ledgerTree,
+      lastExec,
+      ledgerStorage,
+      blockStorage
+    )
+
+    Resource.liftF(service)
+  }
+
+}
--- a/metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
+++ b/metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
@@ -1,0 +1,405 @@
+package io.iohk.metronome.checkpointing.service
+
+import cats.data.NonEmptyList
+import cats.effect.Resource
+import cats.effect.concurrent.Ref
+import cats.implicits._
+import io.iohk.metronome.checkpointing.CheckpointingAgreement
+import io.iohk.metronome.checkpointing.models.Block.{Hash, Header}
+import io.iohk.metronome.checkpointing.models.{
+  ArbitraryInstances,
+  Block,
+  Ledger
+}
+import io.iohk.metronome.checkpointing.service.CheckpointingService.LedgerTree
+import io.iohk.metronome.checkpointing.service.storage.LedgerStorage
+import io.iohk.metronome.checkpointing.service.storage.LedgerStorageProps.{
+  neverUsedCodec,
+  Namespace => LedgerNamespace
+}
+import io.iohk.metronome.hotstuff.consensus.basic.Phase.Commit
+import io.iohk.metronome.hotstuff.consensus.basic.QuorumCertificate
+import io.iohk.metronome.hotstuff.service.storage.BlockStorage
+import io.iohk.metronome.hotstuff.service.storage.BlockStorageProps.{
+  Namespace => BlockNamespace
+}
+import io.iohk.metronome.storage.{
+  InMemoryKVStore,
+  KVCollection,
+  KVStoreRunner,
+  KVStoreState,
+  KVTree
+}
+import monix.eval.Task
+import monix.execution.Scheduler
+import org.scalacheck.Arbitrary.arbitrary
+import org.scalacheck.Prop.{all, classify, forAll, forAllNoShrink, propBoolean}
+import org.scalacheck.{Gen, Prop, Properties}
+
+import scala.concurrent.duration._
+import scala.util.Random
+
+class CheckpointingServiceProps extends Properties("CheckpointingService") {
+
+  type Namespace = String
+
+  case class TestResources(
+      checkpointingService: CheckpointingService[Task, Namespace],
+      ledgerStorage: LedgerStorage[Namespace],
+      blockStorage: BlockStorage[Namespace, CheckpointingAgreement],
+      store: KVStoreRunner[Task, Namespace],
+      ledgerTreeRef: Ref[Task, LedgerTree]
+  )
+
+  case class TestFixture(
+      initialBlock: Block,
+      initialLedger: Ledger,
+      batch: List[Block],
+      commitQC: QuorumCertificate[CheckpointingAgreement]
+  ) {
+    val resources: Resource[Task, TestResources] = {
+      val ledgerStorage =
+        new LedgerStorage[Namespace](
+          new KVCollection[Namespace, Ledger.Hash, Ledger](
+            LedgerNamespace.Ledgers
+          ),
+          LedgerNamespace.LedgerMeta,
+          maxHistorySize = 10
+        )
+
+      val blockStorage = new BlockStorage[Namespace, CheckpointingAgreement](
+        new KVCollection[Namespace, Block.Hash, Block](BlockNamespace.Blocks),
+        new KVCollection[Namespace, Block.Hash, KVTree.NodeMeta[Hash]](
+          BlockNamespace.BlockMetas
+        ),
+        new KVCollection[Namespace, Block.Hash, Set[Block.Hash]](
+          BlockNamespace.BlockToChildren
+        )
+      )
+
+      implicit val store = InMemoryKVStore[Task, Namespace](
+        Ref.unsafe[Task, KVStoreState[Namespace]#Store](Map.empty)
+      )
+
+      Resource.liftF {
+        for {
+          _ <- store.runReadWrite {
+            ledgerStorage.put(initialLedger.hash, initialLedger) >>
+              blockStorage.put(initialBlock)
+          }
+
+          ledgerTree <- Ref.of[Task, LedgerTree](
+            Map(initialBlock.hash -> (initialLedger, initialBlock.header))
+          )
+          lastExec <- Ref.of[Task, Header](initialBlock.header)
+
+          service = new CheckpointingService[Task, Namespace](
+            ledgerTree,
+            lastExec,
+            ledgerStorage,
+            blockStorage
+          )
+
+        } yield TestResources(
+          service,
+          ledgerStorage,
+          blockStorage,
+          store,
+          ledgerTree
+        )
+      }
+    }
+
+    // not used in the impl so a senseless value
+    val commitPath = NonEmptyList.one(initialBlock.header.parentHash)
+
+    val allTransactions = batch.flatMap(_.body.transactions)
+    val finalLedger     = initialLedger.update(allTransactions)
+  }
+
+  object TestFixture {
+    import ArbitraryInstances._
+
+    def gen(minChain: Int = 1): Gen[TestFixture] = {
+      for {
+        block <- arbitrary[Block]
+        ledger = Ledger.empty.update(block.body.transactions)
+        batch    <- genBlockChain(block, ledger, min = minChain)
+        commitQC <- genCommitQC(batch.last)
+      } yield TestFixture(block, ledger, batch, commitQC)
+    }
+
+    def genBlockChain(
+        parent: Block,
+        initialLedger: Ledger,
+        min: Int = 1,
+        max: Int = 6
+    ): Gen[List[Block]] = {
+      for {
+        n      <- Gen.choose(min, max)
+        blocks <- Gen.listOfN(n, arbitrary[Block])
+      } yield {
+        def link(
+            parent: Block,
+            prevLedger: Ledger,
+            chain: List[Block]
+        ): List[Block] = chain match {
+          case b :: bs =>
+            val nextLedger = prevLedger.update(b.body.transactions)
+            val header = b.header.copy(
+              parentHash = parent.hash,
+              height = parent.header.height + 1,
+              postStateHash = nextLedger.hash
+            )
+            val linked = Block.makeUnsafe(header, b.body)
+            linked :: link(linked, nextLedger, bs)
+          case Nil =>
+            Nil
+        }
+
+        link(parent, initialLedger, blocks)
+      }
+    }
+
+    def genCommitQC(
+        block: Block
+    ): Gen[QuorumCertificate[CheckpointingAgreement]] =
+      arbitrary[QuorumCertificate[CheckpointingAgreement]].map {
+        _.copy[CheckpointingAgreement](phase = Commit, blockHash = block.hash)
+      }
+  }
+
+  def run(fixture: TestFixture)(test: TestResources => Task[Prop]): Prop = {
+    import Scheduler.Implicits.global
+
+    fixture.resources.use(test).runSyncUnsafe(timeout = 5.seconds)
+  }
+
+  property("normal execution") = forAll(TestFixture.gen()) { fixture =>
+    run(fixture) { res =>
+      import fixture._
+      import res._
+
+      val execution = batch
+        .map(checkpointingService.executeBlock(_, commitQC, commitPath))
+        .sequence
+
+      val ledgerStorageCheck = store.runReadOnly {
+        ledgerStorage.get(finalLedger.hash)
+      }
+
+      for {
+        results         <- execution
+        persistedLedger <- ledgerStorageCheck
+        ledgerTree      <- ledgerTreeRef.get
+      } yield {
+        val ledgerTreeUpdated = ledgerTree == Map(
+          batch.last.hash -> (finalLedger, batch.last.header)
+        )
+
+        all(
+          "execution successful" |: results.reduce(_ && _),
+          "ledger persisted" |: persistedLedger.contains(finalLedger),
+          "ledgerTree updated" |: ledgerTreeUpdated
+        )
+      }
+    }
+  }
+
+  property("failed execution - no parent") =
+    forAll(TestFixture.gen(minChain = 2)) { fixture =>
+      run(fixture) { res =>
+        import fixture._
+        import res._
+
+        // parent block or its state is not saved so this must fail
+        val execution = batch.tail
+          .map(checkpointingService.executeBlock(_, commitQC, commitPath))
+          .sequence
+
+        execution.attempt.map {
+          case Left(ex: IllegalStateException) =>
+            ex.getMessage.contains("Could not execute block")
+          case _ => false
+        }
+      }
+    }
+
+  property("failed execution - height below last executed") =
+    forAll(TestFixture.gen(minChain = 2)) { fixture =>
+      run(fixture) { res =>
+        import fixture._
+        import res._
+
+        val execution = batch
+          .map(checkpointingService.executeBlock(_, commitQC, commitPath))
+          .sequence
+
+        // repeated execution must fail because we're trying to execute a block of lower height
+        // than the last executed block
+        execution >>
+          execution.attempt.map {
+            case Left(ex: IllegalStateException) =>
+              ex.getMessage.contains("Could not execute block")
+            case _ => false
+          }
+      }
+    }
+
+  //TODO: Validate transactions PM-3131/3132
+  //      use a mocked interpreter client that always evaluates blocks as valid
+  property("parallel validation") = forAll(TestFixture.gen(minChain = 4)) {
+    fixture =>
+      run(fixture) { res =>
+        import fixture._
+        import res._
+
+        // validation in random order so blocks need to be persisted first
+        val persistBlocks = store.runReadWrite {
+          batch.map(blockStorage.put).sequence
+        }
+
+        def validation(
+            validating: Ref[Task, Boolean],
+            achievedPar: Ref[Task, Boolean]
+        ) =
+          Task.parSequence {
+            Random
+              .shuffle(batch)
+              .map(b =>
+                for {
+                  v <- validating.getAndSet(true)
+                  _ <- achievedPar.update(_ || v)
+                  r <- checkpointingService.validateBlock(b)
+                  _ <- validating.set(false)
+                } yield r.getOrElse(false)
+              )
+          }
+
+        for {
+          _ <- persistBlocks
+
+          // used to make sure that parallelism was achieved
+          validating  <- Ref[Task].of(false)
+          achievedPar <- Ref[Task].of(false)
+
+          result     <- validation(validating, achievedPar)
+          par        <- achievedPar.get
+          ledgerTree <- ledgerTreeRef.get
+        } yield {
+          val ledgerTreeUpdated = batch.forall(b => ledgerTree.contains(b.hash))
+
+          classify(par, "parallelism achieved") {
+            all(
+              "validation successful" |: result.forall(identity),
+              "ledgerTree updated" |: ledgerTreeUpdated
+            )
+          }
+        }
+      }
+  }
+
+  //TODO: Validate transactions PM-3131/3132
+  //      use a mocked interpreter client that always evaluates blocks as valid
+  property("execution parallel to validation") = forAllNoShrink {
+    for {
+      f   <- TestFixture.gen(minChain = 4)
+      ext <- TestFixture.genBlockChain(f.batch.last, f.finalLedger)
+    } yield (f, f.batch ++ ext)
+  } { case (fixture, validationBatch) =>
+    run(fixture) { res =>
+      import fixture._
+      import res._
+
+      // validation in random order so blocks need to be persisted first
+      val persistBlocks = store.runReadWrite {
+        validationBatch.map(blockStorage.put).sequence
+      }
+
+      def validation(
+          validating: Ref[Task, Boolean],
+          executing: Ref[Task, Boolean],
+          achievedPar: Ref[Task, Boolean]
+      ) =
+        Random
+          .shuffle(validationBatch)
+          .map(b =>
+            for {
+              _ <- validating.set(true)
+              e <- executing.get
+              _ <- achievedPar.update(_ || e)
+              r <- checkpointingService.validateBlock(b)
+              _ <- validating.set(false)
+            } yield (r.getOrElse(false), b.header.height)
+          )
+          .sequence
+
+      def execution(
+          validating: Ref[Task, Boolean],
+          executing: Ref[Task, Boolean],
+          achievedPar: Ref[Task, Boolean]
+      ) =
+        batch
+          .map(b =>
+            for {
+              _ <- executing.set(true)
+              v <- validating.get
+              _ <- achievedPar.update(_ || v)
+              r <- checkpointingService.executeBlock(b, commitQC, commitPath)
+              _ <- executing.set(false)
+            } yield r
+          )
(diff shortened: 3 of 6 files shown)


metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,186 @@
...
+        _          <- ledgerUpdateSemaphore.releaseN(maxConcurrentUpdates)
+        _          <- pushCheckpoint(checkpoint)
+      } yield ()
+    }
+  }
+
+  private def updateLedger(block: Block): F[Ledger] = {
+
+    // I'm worried that this may loop for a long time (until it reaches a pruned block)
+    // when the initial block is an ancestor of the last committed block - could that happen?
```
aakoshh, line 76: Difficult to say.. I'm trying to get rid of a High Q.C. that has an earlier view than our own Commit Q.C. in the `ConsensusService`. I maintained the possibility in the `RobotService` and checked first if there's an ancestry relationship between the last executed block and the one I'm evaluating. 

I suppose it is possible because of concurrency: you may be asked to evaluate some other proposal block, but it blocks on the semaphore, and by the time it is allowed to go, another one fiber cleared the ledger tree, including the ancestors of this alternative proposal, which will now keep looping.

If you persisted the ledgers of all ancestor blocks, and also looked for it here as a fallback from the tree, wouldn't it feel you with more confidence that it will not loop back to genesis (and hold all blocks in memory along the way)?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,186 @@
...
+  private def getLedger(hash: Block.Hash): F[Option[Ledger]] =
+    ledgerTree.get.map(_.get(hash).map(_._1))
+
+  private def saveCommittedLedger(ledger: Ledger): F[Unit] =
+    storeRunner.runReadWrite {
+      ledgerStorage.put(ledger)
+    }
+
+  private def resetLedgerTree(ledger: Ledger, block: Block): F[Unit] =
+    ledgerTree.set(Map(block.hash -> (ledger, block.header)))
```
aakoshh, line 124: Hm, this seems easy :) I haven't thought about it, always thought in terms of pruning based on ancestry. But it seems true that the commit will always belong to a tip.

aakoshh (commented):
Some initial comments.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: > If you persisted the ledgers of all ancestor blocks, and also looked for it here as a fallback from the tree

Sorry, I was wrong there, since the block to validate can skip some levels. 

I guess one thing that could help is to add a `height` field to the blocks, which could be used to eliminate traversing back more than our last executed block: if the block we are traversing has a lower height, we should have hit the root already, or we are on a different branch.

Not sure if this is important enough to be added to the `Block` type class itself and be supported in `BlockStorage`, but you can definitely do it in the `checkpointing.model.Block` and make use of it here.

aakoshh: A side note: I think the lightest touch to add support for early stop to `BlockStorage.getPathFromAncestor` is to add a `maxDistance` parameter to the method. Since it's not retrieving blocks it cannot check the height, but if on the call site we know the height of our potential ancestor (the root of the state tree) and our new block, we know that if we had to go higher more than that, we should have found a path, or there isn't one.

aakoshh: Added `maxDistance` in https://github.com/input-output-hk/metronome/pull/39/commits/907bbe32432865c5153004ff5abdb27b2da19d72 just in case.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,186 @@
...
+    require(commitQC.phase == Phase.Commit, "Commit QC required")
+    if (commitQC.blockHash != block.hash) {
+      //Several options here, depending on how we want to schedule the execution (BlockExecutor):
+      // - disallow this condition
+      // - ignore
+      // - execute block without constructing CC
+      ().pure[F]
+    } else {
+      for {
+        _          <- ledgerUpdateSemaphore.acquireN(maxConcurrentUpdates)
```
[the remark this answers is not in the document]
aakoshh: I guess there are two reasons for locking: to avoid executing the same block twice for validation reasons (but this is a cheap in-memory operation) and, more importantly, to avoid re-entering blocks into the cache when they are no longer on the longest committed path. 

I thought the second case could also be protected against if, at the time of saving the result into the tree, we checked that it still has the parent in the tree. If it's been cleared out, then we don't need its descendants either.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,186 @@
...
+        ledger     <- updateLedger(block)
+        checkpoint <- constructCheckpoint(ledger, commitQC)
```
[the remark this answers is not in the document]
rtkaczyk: See #54 where the checkpoint is created

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: I've added block height and I'm using it not to traverse beyond last committed/executed block height.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
rtkaczyk: > at the time of saving the result into the tree, we checked that it still has the parent in the tree. If it's been cleared out, then we don't need its descendants either.

Possibly that could get rid off locking. I think I'd just need to change `updateLedger` so that it reads the `ledgerTree` map only once for the whole traversal (instead of for each step). WDYT?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: > Not a problem, just wanted to bring it up that when we see a Commit QC here, it may not be the tip any more, thanks to concurrency.

If we will receive a Commit QC for an ancestor of the last executed block then the `updateLedger` method should raise an error, now that it's looking to block heights

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Ledger.scala
```diff
@@ -21,8 +19,8 @@ case class Ledger(
     * by this point, so we know for example that the new checkpoint is
     * a valid extension of the previous one.
     */
-  def update(transaction: Validated[Transaction]): Ledger =
-    (transaction: Transaction) match {
+  def update(transaction: Transaction): Ledger =
+    transaction match {
```
aakoshh, line 23: I suppose the docstrings make this clear enough.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,208 @@
...
+        .bracket {
+          ledgerUpdateSemaphore.acquireN(maxConcurrentUpdates)
+        } { _ =>
+          for {
+            lastExec <- lastExecutedHeader.get
+            ledger   <- updateLedger(block, lastExec.height)
+            chkpOpt  <- constructCheckpoint(ledger, commitQC)
+            _        <- commitState(block.header, ledger)
+            -        <- chkpOpt.map(pushCheckpoint).getOrElse(().pure[F])
+          } yield false
```
aakoshh, line 77:
```suggestion
          } yield true
```

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,208 @@
...
+  private def getBlock(hash: Block.Hash): F[Option[Block]] =
+    storeRunner.runReadOnly(blockStorage.get(hash))
+
+  private def getLedger(hash: Block.Hash): F[Option[Ledger]] =
+    ledgerTree.get.map(_.get(hash).map(_._1))
+
+  private def commitState(header: Block.Header, ledger: Ledger): F[Unit] = {
+    storeRunner.runReadWrite {
+      ledgerStorage.put(ledger) >>
+        viewStateStorage.setLastExecutedBlockHash(header.hash)
```
aakoshh, line 140: It would be best not to modify the last executed block hash here:
1. You will return a flag to indicate whether this should be done by the executor
2. The state sync mechanism may have done it in the meantime, which is something the executor will check before modifying it

aakoshh (commented):
Looks a bit complicated but doing the right thing I think :)

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: I did not consider state sync. But I wonder isn't it the more reason to update the last executed hash here? We could also use `compareAndSetLastExecutedBlockHash` here and if we detect that it has changed outside of the execution we would update the `ledgerTree` accordingly. This would be just an optimisation though, the CheckpointingService should do fine if the internal in-memory state lags behind.

rtkaczyk: Ah, I guess `CheckpointingService#syncState` will handle that anyway, right?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,208 @@
...
+)(implicit storeRunner: KVStoreRunner[F, N])
+    extends ApplicationService[F, CheckpointingAgreement] {
+
+  override def createBlock(
+      highQC: QuorumCertificate[CheckpointingAgreement]
+  ): F[Option[Block]] = ???
+
+  override def validateBlock(block: Block): F[Option[Boolean]] = {
+    for {
+      lastExec <- lastExecutedHeader.get
```
aakoshh, line 41: This should probably be within the semaphore: in case a commit moves the last executed header, this value that we read before trying to take out the semaphore could be obsolete.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,208 @@
...
+      commitQC: QuorumCertificate[CheckpointingAgreement],
+      commitPath: NonEmptyList[Block.Hash]
+  ): F[Boolean] = {
+    require(commitQC.phase == Phase.Commit, "Commit QC required")
+
+    if (commitQC.blockHash != block.hash) {
+      lastExecutedHeader.get
+        .flatMap { h =>
+          ledgerUpdateSemaphore.withPermit(
+            updateLedger(block, h.height)
```
aakoshh, line 62: Same [comment](https://github.com/input-output-hk/metronome/pull/45/files#r651103786) about the `lastExecutedHeader` being outside the lock.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,208 @@
...
+      Sync[F]
+        .bracket {
+          ledgerUpdateSemaphore.acquireN(maxConcurrentUpdates)
+        } { _ =>
+          for {
+            lastExec <- lastExecutedHeader.get
+            ledger   <- updateLedger(block, lastExec.height)
+            chkpOpt  <- constructCheckpoint(ledger, commitQC)
+            _        <- commitState(block.header, ledger)
+            -        <- chkpOpt.map(pushCheckpoint).getOrElse(().pure[F])
```
aakoshh, line 76: This could be outside the lock, since it's not needed to protect the state any more and does some IO.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
aakoshh: I admit I find it difficult to think though the locking. Currently you allow 4 validations/creations that can all add to the state, but require exclusive access for execution. So there is a theoretical potential for double entry because if we have a chain like this:
```
        D
      / 
A - B 
      \ 
       C
```
And we try to validate `C` and `D` at the same time, and both realise `B` is not in the ledger, they will both concurrently fetch the state of `A`, apply `B` and enter it into the state. So that's not a reason to limit concurrency then.

That leaves the question of removing data from the state and not re-entering it. 

This is a bit tricky because we could commit an interim value when you already have calculated the state for descendants. I would use locking there to check that any state I re-enter as a result of the traversal is not older than the potentially updated last executed state. Not that it would be a big deal if you re-entered either since you clear out everything every time.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: Yes, the `CheckpointingService.syncState` would be called when it's being asked to jump ahead and it can re-establish a correct `ledgerTree`. 

The reason I think to keep it out of here is that if you had to explain to someone how to implement an `ApplicationService` it would be best not to have to explain things like "we also expect you to update this storage when X and Y happens". i thought it was best to make these requests part of the return values from functions, so we can take care of them in the common parts.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,208 @@
...
+      getLedger(block.header.parentHash).flatMap {
+        case Some(ledger) =>
+          updateLedgerByBlock(ledger, block)
+
+        case None if block.header.height <= executedHeight =>
+          Sync[F].raiseError(
+            new IllegalStateException(
+              "Attempting to execute a branch that does not extend last executed block. Aborting execution"
+            )
+          )
```
aakoshh, line 100: It is possible that by the time we try to evaluate the validity of a block, the execution went down a different path. Or that the jump synchronizer cleared out the ancestor block. 

In these cases you could just return `None`, as in you cannot produce the ledger that belongs to that block, which I reckon is only problematic during block execution. Validation and block creation could handle this issue without raising an exception. Or at least validation shouldn't throw an error.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: True, what I meant is that you may be discarding states from the tree that you will still need. 

Say we had this chain where `*` is the first, `**` is the second and `***` is the third commit. When the executor is asked to execute blocks `[C,D]`, it might happen that while `C` is executing, we have already received a commit for `E` and you are being asked to validate `F`. So you construct the ledger for `D`, `E` and `F` in the tree, and _then_ finally get to actually execute `D` because `C` was slow.
```
A - B* - C - D** - E*** - F
```
At that point your code will clear out all content from the tree apart from the ledger that belonged to `D`, and when it comes to execute `E` it will reconstruct it again. 

That's what I was trying to say with the descendant stuff, that if we extracted a generic `KVTreeStorage` then it could say after execution to "prune anything that doesn't descend from D" and keep D, E and F.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: Good change! 

I thought my earlier attempt with `maxDistance` was a bit lacklustre; added `height` to the `Block` type class in https://github.com/input-output-hk/metronome/pull/55

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: To put my money where my mouth is I extracted a `KVTree` in https://github.com/input-output-hk/metronome/pull/56

Up to you if you want to use it or not (with an in-memory store, like the `BlockSynchronizer`), but at least it's available.

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
```diff
@@ -0,0 +1,407 @@
...
+        results         <- execution
+        persistedLedger <- ledgerStorageCheck
+        ledgerTree      <- ledgerTreeRef.get
+      } yield {
+        val ledgerTreeUpdated = ledgerTree == Map(
+          batch.last.hash -> (finalLedger, batch.last.header)
+        )
+
+        all(
+          "executed correctly" |: results.reduce(_ && _),
```
aakoshh, line 201: By "executed correctly" you mean that "executed according to expectations", which is that all executions returned `true`, not that a `false` would signal incorrect execution, right? 

Just wondering if something like "persisted all executions" would make it clearer.

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
```diff
@@ -0,0 +1,407 @@
...
+        import res._
+
+        // parent block or its state is not saved so this must fail
+        val execution = batch.tail
+          .map(checkpointingService.executeBlock(_, commitQC, commitPath))
+          .sequence
+
+        execution
+          .onErrorHandle(_.isInstanceOf[IllegalStateException])
+          .map(_ == true)
```
aakoshh, line 222: I was starting at this for a while to understand why the last `map(_ == true)` is necessary. Maybe something like this would be clearer:
```scala
execution.attempt.map {
  case Left(ex: IllegalStateException) => ex.getMessage.contains("Could not execute block")
  case _ => false
}
```

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
```diff
@@ -0,0 +1,407 @@
...
+          batch.map(blockStorage.put).sequence
+        }
+
+        def validation(
+            validating: Ref[Task, Boolean],
+            achievedPar: Ref[Task, Boolean]
+        ) =
+          Task.parSequence {
+            Random
+              .shuffle(batch)
```
aakoshh, line 264: You can make a seed in `forAll` and use it to create a `Random` instance, that way even the shuffle would be repeatable, in case something fails.

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
```diff
@@ -0,0 +1,407 @@
...
+        validationBatch.map(blockStorage.put).sequence
+      }
+
+      def validation(
+          validating: Ref[Task, Boolean],
+          executing: Ref[Task, Boolean],
+          achievedPar: Ref[Task, Boolean]
+      ) =
+        Random
+          .shuffle(validationBatch)
```
aakoshh, line 320: Same thing as above with the seed.

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
```diff
@@ -0,0 +1,407 @@
...
+          .map(b =>
+            Task.defer {
```
aakoshh, line 322: What's the extra `defer` for?

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
```diff
@@ -0,0 +1,407 @@
...
+          "parallelism achieved" |: par,
+          "validation successful" |: validationsAfterExec.forall(identity),
+          "execution successful" |: executionRes.forall(identity),
+          "ledger persisted" |: persistedLedger.contains(finalLedger),
+          "ledgerTree reset" |: ledgerTreeReset
+          //TODO: possible optimisation
+          //  This test shows that upon committing a block the projected state of its descendants may be lost.
+          //  It's not techically wrong, as blocks will still validated correctly, but that state could potentially
+          //  be reused
+          //"ledgerTree contains validations" |: validationsSaved
```
aakoshh, line 401: Good to have a comment and test case to show this :heart:

aakoshh (commented):
This looks great! Much easier to think it through without the locks. Love the tests too, good job!

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: When I initially checked the parallelism by adding simple `println`s (within the `flatMap` bodies inside the methods) I found out that all validation are run before execution, as if the Tasks were actually being run upon creation. `Task.defer` seemed to rectify this behaviour. I also added the parallelism checks to make sure.

However now, I am not able to reproduce that behaviour, even when I just do `.map(b => checkpointingService.validateBlock(b)) //or execute`. I wish I could explain this, but all I can do is remove those `Task.defer`.

Regarding the parallelism checks though. We've had a [failure](https://app.circleci.com/pipelines/github/input-output-hk/metronome/471/workflows/e9b3767d-d3e6-4b23-94a6-17b5f02bd286/jobs/527). We've had a [success](https://app.circleci.com/pipelines/github/input-output-hk/metronome/473/workflows/99647d7b-3fc9-4894-b53d-365bcfa475e5/jobs/529), too. On a single occasion, I also observed a failure locally. I guess one way to improve this would be to increase minimum batch sizes (more opportunity for parallelism), but it may be that these tests will always be flaky, especially if running on a single-thread scheduler. What do you think? Should we keep them?

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: I actually forgot to finish this comment 🙃 
I meant to note that a possible improvement is, upon `saveState`, not to remove any descendants (if found) of the committed block. Do you think we should do that?

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: Did that in #54

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: Did that in #54

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,200 @@
...
+      storeRunner: KVStoreRunner[F, N]
+  ): Resource[F, CheckpointingService[F, N]] = {
+    val lastExecuted: F[(Block, Ledger)] =
+      storeRunner.runReadOnly {
+        val query = for {
+          blockHash <- OptionT.liftF(
+            viewStateStorage.getLastExecutedBlockHash
+          )
+          block <- OptionT(blockStorage.get(blockHash))
+          //TODO: have a genesis (empty) state in ledger storage
```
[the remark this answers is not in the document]
rtkaczyk: Cool! Should I still keep this comment?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,200 @@
...
+        ledgers.get(block.header.parentHash) match {
+          case Some((oldLedger, _)) =>
+            doUpdate(oldLedger)
+
+          case None if block.header.height <= execHeight =>
+            OptionT.none
+
+          case None =>
+            for {
+              //TODO: no reason to look up storage for execution; validation???
```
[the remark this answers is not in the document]
rtkaczyk: Yeah, it was just meant for myself to start the discussion 😄 

The way that `BlockExecutor` works it shouldn't be necessary to ever look up block storage when running `executeBlock`. What about validation though?

Blocks that are behind the last committed block will be rejected based on height anyway. But perhaps it may happen that we will be validating a descendant of last committed block whose parent has not been validated/executed before. In such case it would make sense to keep the storage look-up, but I'm curious about your input.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,200 @@
...
+            //TODO: PM-3110:
+            // chkpOpt <- constructCheckpoint(ledger, commitQC)
+            _ <- saveState(block.header, ledger)
+            //TODO: PM-3110:
+            // _ <- chkpOpt.map(pushCheckpoint).getOrElse(().pure[F])
+          } yield true
+
+      case None =>
+        Sync[F].raiseError(
+          //TODO: better error info with Either?
```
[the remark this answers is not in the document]
rtkaczyk: Yeah, it would only be helpful if we mess up something badly. I guess we can always add some debugging code in that case. I will remove this comment

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,200 @@
...
+    extends ApplicationService[F, CheckpointingAgreement] {
+
+  override def createBlock(
+      highQC: QuorumCertificate[CheckpointingAgreement]
+  ): F[Option[Block]] = ???
+
+  override def validateBlock(block: Block): F[Option[Boolean]] = {
+    OptionT(updateLedger(block)).flatMapF { ledger =>
+      val hashValid = ledger.hash == block.header.postStateHash
+      validateTransactions(block.body, ledger).map {
```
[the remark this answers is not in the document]
rtkaczyk: Ok, I misunderstood that.

So I could just read the `ledgerTree` ref for the parent state. The problem is that if for some reason we validate the last committed block (after we did `executeBlock` on it) then parent state will be gone and the validation will fail. Do you think that's acceptable? 

Alternatively I  guess I could make `updateLedger` return the parent state as well.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: No lock anymore

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: No longer relevant

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: No longer relevant

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: What would be the benefit of using `KVTree` instead?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
rtkaczyk: I don't this is relevant anymore with locking gone. If so, please resolve this convo.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,186 @@
...
+      commitQC: QuorumCertificate[CheckpointingAgreement],
+      commitPath: NonEmptyList[Block.Hash]
+  ): F[Unit] = {
+    require(commitQC.phase == Phase.Commit, "Commit QC required")
+    if (commitQC.blockHash != block.hash) {
+      //Several options here, depending on how we want to schedule the execution (BlockExecutor):
+      // - disallow this condition
+      // - ignore
+      // - execute block without constructing CC
+      ().pure[F]
```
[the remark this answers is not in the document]
rtkaczyk: Yes, we are computing the possible checkpoint proof here now. For a moment I though we could let the `BlockExecutor` save the block as _last executed_, but it raises some complications with checkpoint certificate creation. I will comment more on it in #54

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,200 @@
...
+    require(commitQC.phase == Phase.Commit, "Commit QC required")
+    updateLedger(block).flatMap {
+      case Some(ledger) =>
+        if (block.hash != commitQC.blockHash)
+          saveState(block.header, ledger).as(true)
+        else
+          for {
+            //TODO: PM-3110:
+            // chkpOpt <- constructCheckpoint(ledger, commitQC)
+            _ <- saveState(block.header, ledger)
```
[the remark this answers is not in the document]
rtkaczyk: Yes, I did not realise this for a moment. In #54 I bring the previous behaviour back. Let's discuss more there

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: There is a separate check for persisting the state. Maybe simply "executed as expected"?

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
aakoshh: Sometimes I put in a short `sleep` to simulate time and raise the chances of things actually happening concurrently. While you could do that in this case it would probably not be inside a critical section, where it could cause any issues. 

What I'd try is to [classify](https://github.com/typelevel/scalacheck/blob/main/doc/UserGuide.md#collecting-generated-test-data) tests by whether they achieved parallelism or not. That would give us the confidence that the test is written correctly, even if occasionally it's not actually concurrent on CI.

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
aakoshh: I would have personally tried to go that way, hence the generalisation of the tree storage in https://github.com/input-output-hk/metronome/pull/56, but I also like the simple assurance against memory leaks that a full reset provides. 

In practice it's hopefully not going to be the case that we have a long chain of uncommitted blocks building up that have to be re-entered into the ledger state all the time, but this optimisation seemed like something the flagship use case would do.

The important thing is that now we have these tests, so you refactoring can be done with confidence.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
aakoshh: I'd keep it without a `TODO`, rather as a note that we expect this constructor expects that the genesis state is inserted by the application composer. It's something that I think I forgot to do in the robot tests, and then wondered where I got the exception from. A reminder could help.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
aakoshh: So you are saying that when the class is created, you do 1 lookup in persistent ledger storage, put it in the in-memory ledger tree, then executions just build on that one, and only do writes to the persistent ledger store. You are wondering if we would ever need to read the persistent state store for validation. 

>  But perhaps it may happen that we will be validating a descendant of last committed block whose parent has not been validated/executed before.

It is indeed possible that we miss a (few) round(s) for any reason, and we need to validate a block that is the descendant of the block we know about as last committed, but we have some missing in the chain. In this case the `SyncService` would download these but not call validation on them, as their validity is proven by the High Q.C. extended by the proposed block we were given. The missing ancestor blocks, up to the High Q.C. would be persisted, and `validateBlock` called on the proposed child. 

Your algorithm should walk up the tree until it finds the state of the last committed block in memory. I don't see where it could find anything in persistent storage along the way that it doesn't have in memory. 

Are we talking about the same thing?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
aakoshh: Tracing is a nice way to add as much debug content as you need.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
aakoshh: Right, you mean that `updateLedger` goes back until it finds a state in memory, then builds up returns the last ledger, while inserting it the interim ledgers into memory; at the same time all of it could be removed by the execution. And because the execution removes all the descendants, this can cause a race condition, and if you just tried the read the parent state from the tree, it might not be there any more. 

No, I don't think this should be a reason to fail validation, since we know that this is just a quirk of the current implementation, just incidental complexity, not inherent to the problem.

In the robot example I had a `projectState(block.parentBlockHash)` construct, which was common to `validateBlock`, `executeBlock` and `createBlock` as well, so the state by parent hash is kind of a recurring theme.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: Well, you have a `ledgerTree`, which is a map of ledger and header indexed by block hash; while you can traverse upwards because the headers show you the parent, it's not easy to go in the other direction, towards the descendants. As you noted in tests, `executeBlock` removes all other entries from the tree, potentially causing re-evaluation of the update rules, repeated block retrievals. You identified as a possible optimisation that these states could be kept in the map. But that requires more coding to select which blocks are descendants. 

This is where the `KVTree` could help, since it already has methods to implement this kind of pruning, so you don't have to do it again. It's still not a panacea from the concurrency perspective, as I myself had to be careful while maintaining the in-memory block tree in `BlockSynchronizer` to make sure I don't end up with a forest. That's why I'm not trying to push it on you, just wanted to say that if that optimisation was troublesome, this could be a way to reuse what we already have.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
rtkaczyk: > No, I don't think this should be a reason to fail validation, since we know that this is just a quirk

The behaviour already is that the ancestors of the last committed block will fail validation even if they were correctly executed before, which is due to the [height condition](https://github.com/input-output-hk/metronome/blob/1a3f15eb976f58996f5249b52991d8585ac81380/metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala#L100). Because the last committed block is also set in stone, I think it's fine to extend this behaviour also to this block.

If you think however that validating the ancestors should always succeed, then that would be a good reason to fix this. Let me have a look at the robot.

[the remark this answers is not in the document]
rtkaczyk: In the robot example you query the storage quite freely. For efficiency I wanted to avoid that here as much as possible. Do you think that's a misguided optimisation?

Secondly, I was thinking of a possible attack where someone intentionally sends us old blocks to force us to work on reevaluating the state. Hence the height condition. With proper pruning that's not much of a hinderance, but again that would mean that after some threshold validating already executed blocks would fail. So why not, for simplicity, make that threshold equal to the commit?

Going to wait for your input before attempting a fix.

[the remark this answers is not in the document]
rtkaczyk: Btw, do you think the name `projectLedger` would be better than `updateLedger`? I kinda like the verb "project"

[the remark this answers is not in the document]
aakoshh: > I kinda like the verb "project" 

Yay, I named something right :partying_face: For a moment I wasn't sure what to say, because "project" doesn't seem like a final thing, but then I remembered you have `saveState` when you want to persist something. So maybe `projectLedger` and `saveLedger` could be an intuitive duo?

> In the robot example you query the storage quite freely. For efficiency I wanted to avoid that here as much as possible. Do you think that's a misguided optimisation?

No, I think it's a good goal. I wanted to keep the robot stuff simple (close to being obviously correct, no separate unit tests), since it's just an example, but always thought that we would spend more time on checkpointing to get the most out of it.

> I was thinking of a possible attack where someone intentionally sends us old blocks to force us to work on reevaluating the state. Hence the height condition.

I agree, the height condition is very useful; there is a similar check carried out [here](https://github.com/input-output-hk/metronome/blob/pm-3353-quorum-certificate-phase-typed/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala#L97-L101).

> If you think however that validating the ancestors should always succeed, then that would be a good reason to fix this.

Sorry I think you misunderstood. What I meant is that we shouldn't fail to validate in this scenario:
```
       B4
      / ^ 
     /  validate
    /
B0* - B1* - B2 - B3 
      ^          ^
      exec       validate
```

Say we have commits for `*` blocks, we are currently executing B1 while validating B3 and B4.
When the execution of B1 is done, the current code will clear everything else from the `ledgerTree`, so a code like this could fail:

```scala
for {
  ledger <- OptionT(updateLedger(b3))
  parentLedger <- OptionT(ledgerTreeRef.get.map(_.get(b2.blockHash)))
  isValid <- validate(b3.body, parentLedger)
} yield isValid
```

The problem is that by the time we try to read `parentLedger` it may be gone _not_ because the ancestor B2 is not on the main chain but simply because executing `b1` (which has a commit) has purged the state we just calculated. That's what I called a "quirk".

What you described was the situation for B4, where the disappearance of B0 state means B4 is no longer valid. In that situation, you are right, it doesn't matter any more if we don't consider this block valid any more.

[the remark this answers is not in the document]
rtkaczyk: I see, but the thing is b3 will be validated correctly anyway, as noted in the [relevant test](https://github.com/input-output-hk/metronome/blob/eb1df073b664df04e1e5249c5fd017795dd57175/metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala#L399) (well, I'm 90% sure, now that you suggest otherwise).

The reason is that if the `ledgerTree` gets cleared during validation:
- after the ancestor ledger has been read: then we'll simply apply the descendant blocks (which we already have) to it. We just won't be able to save the intermediate ledgers in the tree, which is wasteful.
- before the ancestor ledger has been read: then we'll keep retrieving blocks from storage until we reach committed ledger in the `ledgerTree` (in this case we'll even be able to save the intermediate ledgers in the tree; only the retrieval is suboptimal).

So the issue you bring is not really of correctness, but of optimisation, and will be handled by the smarter descendant-aware clearing of the `ledgerTree`. (It's already done, but I'm looking if `KVTree` makes it easier)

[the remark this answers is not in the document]
rtkaczyk: The point I was making was that in your example `B1` would fail validation, after it has already received a `*`. But I would argue that this is not an issue, as there shouldn't be any reason to process an already committed block in any way. Do you agree?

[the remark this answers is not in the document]
aakoshh: We may be talking past each other now. I'm trying to make a point about this comment of yours: 

> So I could just read the ledgerTree ref for the parent state. The problem is that if for some reason we validate the last committed block (after we did executeBlock on it) then parent state will be gone and the validation will fail. 

The "just read `ledgerTree` ref for the parent state"  part in my snippet was `parentLedger <- OptionT(ledgerTreeRef.get.map(_.get(b2.blockHash)))`. This may return `None` because the execution of an ancestor (B1) of the block being validated (B3) could have removed the parent (B2) state, in which case we can falsely conclude that the block is invalid. 

> but the thing is b3 will be validated correctly anyway, as noted in the relevant test

It is true that this test demonstrates that `executeBlock` and `validateBlock` can be called concurrently safely, however, that's for the current logic of `updateLedger` and `validateBlock`, which is not exactly doing the right thing, as per my comment that opened this discussion: you are projecting the state of B3 in the example, and that's done via `updateLedger` returning the chain of ledgers. But you need the parent state! With that change, if it was implemented in the "just read it after calling updateLedger" way, it could fail in some occasions.

> The point I was making was that in your example B1 would fail validation, after it has already received a *

I didn't think B1 would fail validation as in my example it was already being executed, not validated. I was hypothesising about how B3 could fail validation because of the execution of B1, _if_ the code was changed in certain ways :slightly_smiling_face:

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: I decided against using `KVTree` for 2 reasons:
- setup complexity
- maintaining parent and child links would be difficult to get right in the context of parallel updates

In the current design we either `clearLedgerTree` or only add entries to it. This can be considered thread-safe in our application. By taking advantage of `height` and parent links, pruning the `Map` of non-descendants wasn't too difficult and should be reasonably efficient

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
rtkaczyk: Right, I see what you mean. The improved `clearLedgerTree` solves this problem too. The required parent will be available for validation as long as it's no older than the last committed ledger.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
rtkaczyk: Is that a suggestion for now or something to keep in mind? I did not add any tracing now, as I don't suppose this is likely to surface as an issue.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
rtkaczyk: > Are we talking about the same thing?

Not quite 😉, but I wasn't clear about which storage I was talking about. I meant specifically the `blockStorage` lookup that occurs in `projectLedger`. It is not needed for execution because `BlockExecutor` will take care of providing blocks, but based on your comment above, I believe we are in an agreement that it is still needed to facilitate fully asynchronous validation. Comment removed

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: `classify` is pretty neat, but I would still like to specify a requirement on it, like "require 50% of test cases to be achieve parallelism". Doesn't seem possible though. I added a note to the maintainer to make sure to observe these stats when modifying the test or the service.

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/CheckpointingServiceProps.scala
rtkaczyk: Done, also solves the problem of validating against the parent ledger

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: Sweet, glad we explored all options :+1:

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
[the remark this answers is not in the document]
aakoshh: I only mentioned it in the context of "I guess we can always add some debugging code". 

Exceptions returned by this code are traced in the `BlockExecutor`, but if you ever wanted to add actual debugging, it would be a good indication for the need to add tracing.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -0,0 +1,239 @@
...
+      case None =>
+        Sync[F].raiseError(
+          new IllegalStateException(s"Could not execute block: ${block.hash}")
+        )
+    }
+  }
+
+  private def projectLedger(block: Block): F[Option[Ledger]] = {
+    (for {
+      ledgers    <- ledgerTree.get
```
aakoshh, line 90: To avoid having to choose a non-clashing name for the thing I `.get` from a `Ref`, I adopted a postfix naming convention, so the class level variable would be `ledgerTreeRef: Ref[F, LedgerTree]` and the variable just `ledgerTree`. Not sure it bothered you, just thought I'd mention it.

aakoshh (approved):
Awesome! :champagne:

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Cool, I'll add this to #54

PR #46, opened 2021-06-01
PM-3133: Make Quorum Certificate generic in Phase 

aakoshh:
The result of a PR discussion, positing whether we should have separate `QuorumCertificate` types for each phase, to get rid of some assertions. 

Made `QuorumCertificate` generic in `P <: VotingPhase`, so we can state which kind we must have in certain fields or method parameters.


aakoshh:
Closing in favour of https://github.com/input-output-hk/metronome/pull/47

PR #47, opened 2021-06-01
PM-3353: Make Quorum Certificate generic in Phase

aakoshh:
The result of a PR discussion, positing whether we should have separate QuorumCertificate types for each phase, to get compile time assurance where possible that the right kind are passed around, rather than just rely on assertions.

Tested with the robot integration tests which exercise the database and the network codecs as well.

Removed existing assertions and checks that can now be expected to be validated by the compiler, as well as tests that generated now illegal states.

### DIFF
--- a/metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/basic/QuorumCertificate.scala
+++ b/metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/basic/QuorumCertificate.scala
@@ -13,29 +13,33 @@ case class QuorumCertificate[A <: Agreement, +P <: VotingPhase](
     blockHash: A#Hash,
     signature: GroupSignature[A#PKey, (VotingPhase, ViewNumber, A#Hash), A#GSig]
 ) {
+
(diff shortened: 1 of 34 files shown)


metronome/hotstuff/consensus/src/io/iohk/metronome/hotstuff/consensus/basic/QuorumCertificate.scala
```diff
@@ -2,13 +2,29 @@ package io.iohk.metronome.hotstuff.consensus.basic
...
     blockHash: A#Hash,
     signature: GroupSignature[A#PKey, (VotingPhase, ViewNumber, A#Hash), A#GSig]
-)
+) {
+
+  /** In protocol messages we can treat QCs as `QuorumCertificate[A, VotingPhase]`,
+    * and coerce to a specific type after checking what it is. We can also coerce
+    * back into the supertype, if necessary.
+    */
+  def coerce[V <: VotingPhase](implicit
```
rtkaczyk, line 21: Nice!

rtkaczyk (approved):
Looks better this way

PR #50, opened 2021-06-04
PM-3129: Add LocalConnectionManager on top of the remote one.

aakoshh:
Adds a `LocalConnectionManager` to the `networking` package that we can use to connect to a single address. It reuses the `RemoteConnectionManager` which already takes care of reconnections, and manages the connections whichever side initiated them first. 

The idea is that instead of the Checkpointing Service acting as a client and the PoW interpreter acting as a server, they can both try to connect to each other, and keep reconnecting if the other side disappears for a while. It would be enough if one side does it (the Service probably being a better candidate) but this way we should be able to reuse all our existing code, and if the Checkpointing Service is down, we should see some warnings on the Interpreter side as well.


KonradStaniec (approved):
It seems like neat idea !

PR #51, opened 2021-06-04
PM-3130: Interpreter SDK

aakoshh:
Introduce simple seams for the communication between the Checkpointing Service and the Interpreter.
* `InterpreterRPC`: The RPC subset of the `InterpreterMessages` that the PoW host system has to implement and the Service can call remotely. 
* `ServiceRPC`: The RPC subset of the `InterpreterMessages` that the PoW host system can use to notify the Service about events, such as adding things to the mempool.
* `InterpreterService`: Binds the above two together and starts processing network messages, using the `LocalConnectionManager` from https://github.com/input-output-hk/metronome/pull/50

The `ServiceRPC` methods are best-effort based: if the Service is not available, the notification will be dropped (although we could add retries), assuming that other federation members will achieve the same result by the time our Service instance comes back online. There's tracing so we can at least log that the Service is not available.

The `InterpreterRPC` methods all return optional results: this is so the Interpreter can decide not to respond at all, in case it thinks it would be too expensive to do so, and it's happy to defer to the federation, in case a quorum is reached without its participation. This way the interface can also be used directly on the Service side and reflect the best-effort nature of the communication.

The `InterpreterService` applies a timeout to each request delegated to the `InterpreterRPC`, but that's just to make sure we aren't holding callbacks forever in memory in case the `InterpreterRPC` is faulty and never completes the async responses. The Service is expected to use an `RPCTracker` with its own timeouts. 

We could add more traced events, like "block body created", but I figured everything else is in the control of the `InterpreterRPC`.


metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
```diff
@@ -0,0 +1,35 @@
+package io.iohk.metronome.checkpointing.interpreter
+
+import io.iohk.metronome.checkpointing.models.{Block, Transaction, Ledger}
+import io.iohk.metronome.checkpointing.models.CheckpointCertificate
+
+/** `InterpreterRPC` is the interface that the Service can call on the Interpreter
+  * side to send queries and commands. It provides RPC style methods for some of the
+  * `InterpeterMessage` types, namely the ones `with Request with FromService`.
```
rtkaczyk, line 8: "Service" is an ambiguous word. Since this is a very concrete application I think it's both appropriate and much clearer to refer to the `CheckpointingService`.

"on the Interpreter side" - by that I understand this will be called from an interpreter implementation that lives within the PoW node. But that's not the case, right?

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
```diff
@@ -0,0 +1,161 @@
...
+import io.iohk.metronome.checkpointing.models.Transaction
+import io.iohk.metronome.networking.LocalConnectionManager
+import io.iohk.metronome.tracer.Tracer
+import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
+import io.iohk.metronome.core.messages.RPCPair
+
+/** The `InterpreterService` is to be used on the Interpreter side to
+  * manage the behind-the-scenes messaging with the Service.
+  */
```
rtkaczyk, line 17: Again, I think this description could be clearer. I'm actually lost here. I thought this was meant to be used on the PoW side, but then it depends on the `InterpreterRPC`. Could you clarify?

rtkaczyk: Is it actually meant for both the PoW node and Hotstuff node?

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
```diff
@@ -0,0 +1,161 @@
...
+      localConnectionManager.sendMessage(message).flatMap {
+        case Left(_)  => tracer(ServiceUnavailable(message))
+        case Right(_) => ().pure[F]
+      }
+  }
+
+  /** Start processing Service messages and in the background, delegating to the `interpreterRPC`,
+    * which is the domain specific Interpreter implementation in the host system.
+    *
+    * Returns a `ServiceRPC` instance that the host system can use to send notifications to the Service.
```
rtkaczyk, line 144: What is the "host system"? What is the "Service"?

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
aakoshh: Hm, maybe you are right. These both live in the `checkpointing` package, so "Interpreter" and "Service" always implicitly mean "Checkpointing Interpreter" and "Checkpointing Service" when they are mentioned in the same sentence and capitalised. I didn't want to concretely mean the `CheckpointingService` class, just the Service and the Interpreter in the ]architecture diagram](https://github.com/input-output-hk/metronome/blob/develop/docs/architecture.png) (where Ranvir just called it the "BFT Service" actually, but that's split into two now).

I'm not the best at naming things though. @dzajkowski was it confusing to you as well?


> "on the Interpreter side" - by that I understand this will be called from an interpreter implementation

Maybe not the best of wording then. By "the Service can call on the Interpreter" I meant that this is an RPC interface representing the Interpreter to the Service.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
aakoshh: Yes, I'm guilty as charged abusing the "Service" name, sorry for the confusion. This one is a service in the technical sense as it will actually handle to messages coming from the Checkpointing Service. Okay, you are right, I will add "Checkpointing" to disambiguate. Would be nice to find a different work for the "BFT + Checkpointing ~~Service~~ ???" side.

> I thought this was meant to be used on the PoW side, but then it depends on the `InterpreterRPC`. Could you clarify?

Yeah, so the idea is that the `InterpreterRPC` is an abstract interface, these are the RPC methods that the Interpreter implements. However, that doesn't say it lives on one place or the other: on the Interpreter side there should be a "server/service" that handles the requests, and on the Service side there would be a "client" that sends the requests. 

This is similar to [gRPC](https://cloud.google.com/apis/design/naming_convention#interface_names): we could have a `service Interpreter` with methods like `rpc CreateBlockBody` in the `checkpointing.proto` file, and then you can generate the server and the client side stubs as well. I'm aware that `service Service` would make a bad name if it had to be in gRPC :)

`InterpreterRPC` is passed to `InterpreterService.apply` because that's how for example the Mantis team would inject their logic into the machinery. The return value is a `ServiceRPC` because that's the interface _they_ can call to send messages to the Service.

On the [Service side](https://github.com/input-output-hk/metronome/pull/53), we have the opposite where the `ServiceRPC` implementation is passed to the `InterpreterClient` and an `InterpreterRPC` is returned. 

I can understand that he naming seems arbitrary, since both have client as well as server aspects, being a bi-directional setup. 

Do you have a better suggestion?

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
aakoshh: By the host system I meant that this is a library imported into the PoW node. The Service, as usual, is the BFT+Checkpointing combo. 

Should we just saw "PoW node"?

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
aakoshh: > will be called from an interpreter implementation that lives within the PoW node

As per my [other comment](https://github.com/input-output-hk/metronome/pull/51#discussion_r661281482), the interpreter implementation living inside the PoW node will not call the `InterpeterRPC` but rather be the one that is called - in fact implementing this interface is what it would mean to implement the interpreter. The calling will be done from the Checkpointing Service where a client implementation lives.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
rtkaczyk: Or "host system (ie. the PoW node)" if you think the more generic term can be meaningful in another way.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
rtkaczyk: > on the Interpreter side there should be a "server/service" that handles the requests, and on the Service side there would be a "client" that sends the requests.

Ok, I get it now :+1:

> Do you have a better suggestion?

Only to try to disambiguate by saying "Checkpointing Service". If you keep those words separate, then you're not really directly referring to the class

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
rtkaczyk: I would rephrase that sentence to:

```
/** `InterpreterRPC` is the interface that the Checkpointing Service can call to send queries 
  * and commands to the Interpreter.
```

rtkaczyk: Maybe I would also add something like: 

```
* It is also the interface that the Interpreter implements in order to process the said queries 
* and commands. Thus we have separate client- and server-side implementations.
```

It probably sounds obvious to you, but to me it wasn't :)

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
aakoshh: Thanks, done.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterService.scala
aakoshh: Added "Checkpointing" throughout this class.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
aakoshh: Thanks, added both changes. 

Totally fair that it wasn't obvious. It was inspired by how I did things in discovery, where the RPC interface is implemented both by [DiscoveryNetwork](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryNetwork.scala#L31-L40) and the [DiscoveryService](https://github.com/input-output-hk/scalanet/blob/develop/scalanet/discovery/src/io/iohk/scalanet/discovery/ethereum/v4/DiscoveryService.scala#L112), calling each other. Not sure how well it will work in practice, but this is how the dice fell that way in my head.

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
```diff
@@ -0,0 +1,275 @@
...
+          inside(m2) { case NewProposerBlockRequest(_, txn) =>
+            txn shouldBe pb
+          }
+        }
+    }
+  }
+
+  it should "send messages from the Service to the Interpreter" in test {
+    new Fixture {
+      val mockValidationResult = true
```
rtkaczyk, line 64: Minor improvement, but maybe a random value would give more confidence?

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
```diff
@@ -0,0 +1,275 @@
...
+  ): Resource[Task, LocalConnectionManager[
+    Task,
+    ECPublicKey,
+    InterpreterMessage
+  ]] = {
+    import DefaultInterpreterCodecs.interpreterMessageCodec
+
+    implicit val networkTracers = NetworkTracers(
+      Tracer.noOpTracer[Task, NetworkEvent[ECPublicKey, InterpreterMessage]]
+      // Tracer.instance[Task, NetworkEvent[ECPublicKey, InterpreterMessage]](e =>
```
rtkaczyk, line 234: If it's your intention to keep it around, which may be a good idea, then I'd rather precede it with a note, so that it doesn't look accidental. Something like:

```
// Uncomment below to log network events to stdout
```

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
```diff
@@ -0,0 +1,275 @@
...
+      //   Task(println(e))
+      // )
+    )
+
+    for {
+      implicit0(scheduler: Scheduler) <- Resource.make(
```
rtkaczyk, line 240: Cool, I didn't know about `implicit0`

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
aakoshh: Good idea :+1:

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
aakoshh: It comes from https://github.com/oleg-py/better-monadic-for

Needs the type to be declared.

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
aakoshh: Done.

metronome/checkpointing/interpreter/test/src/io/iohk/metronome/checkpointing/interpreter/InterpreterServiceSpec.scala
aakoshh: Done.

rtkaczyk (approved):
Elegant!

PR #52, opened 2021-06-07
PM-3092: Connection conflict resolution based on ephemeral port ordering

aakoshh:
The integration tests of https://github.com/input-output-hk/metronome/pull/51 failed with often with this kind of network log:

```
localhost/127.0.0.1:34953 - 2021-06-04T16:43:30.044162Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd9e...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:30.050472Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d000...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:30.082489Z: ConnectionDiscarded(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d000f...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:30.127455Z: ConnectionDeregistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d0...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:30.139837Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd9e...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:30.145508Z: ConnectionDeregistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:31.183316Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd9e...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:31.189024Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d000...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:31.281460Z: ConnectionDiscarded(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d000f...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:31.292620Z: ConnectionDeregistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d0...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:31.297326Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd9e...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:31.300318Z: ConnectionDeregistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:31.301845Z: ConnectionSendError(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d000f...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:37597 - 2021-06-04T16:43:32.418044Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x7c308620b5c8d000...)),localhost/127.0.0.1:34953))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:32.421764Z: ConnectionRegistered(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd9e...)),localhost/127.0.0.1:37597))
localhost/127.0.0.1:34953 - 2021-06-04T16:43:32.451721Z: ConnectionDiscarded(Peer(ECPublicKey(ByteVector(64 bytes, 0x697fa1a96f44dd9ef...)),localhost/127.0.0.1:37597))
```

The reason was that both sides opened connections to each other like this (port values fictional):
```
                       Service           Interpreter
------------------------------           ------------------------------
  server listener port - 9871:<-\    /-> :9872 - server listener port
                             |   \  /    |
                             |    \/     |
                             |    /\     |
                             |   /  \    |
ephemeral client port - 54321:--/    \-- :65432 - ephemeral client port
------------------------------           ------------------------------
```
On each side there were two connections: one incoming and one outgoing. Both sides got their outgoing connection registered first, then they had to deal with the incoming one, which was a duplicate connection _from_ the same peer public key they already had a connection _to_. Their conflict resolution logic was symmetrical, so they both chose the same victim, say, their respective outgoing connections, which resulted in both connections being closed, since both were the outgoing connection of somebody. 

The PR adds an additional mechanism based on https://github.com/input-output-hk/scalanet/pull/134, which is that if the connection has only been opened recently, we assume that's short enough time that the reason we see a duplicate is not because the other side is trying to reconnect after the one we have in our registry has failed, but because we opened opposite directions at the same time; in this case both sides use an ordering of connections that gives the same result on each node, based on the ephemeral port used by the other side to initiate the connection from. 

There are no tests in this one, I used the integration tests in the other PR as an acceptance criteria.

### DIFF
--- a/metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
+++ b/metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
@@ -22,13 +22,16 @@ import monix.tail.Iterant
 
 import java.net.InetSocketAddress
 import scala.util.control.NoStackTrace
+import scala.concurrent.duration._
+import java.time.Instant
 
 class ConnectionHandler[F[_]: Concurrent, K, M](
     connectionQueue: ConcurrentQueue[F, ConnectionWithConflictFlag[F, K, M]],
     connectionsRegister: ConnectionsRegister[F, K, M],
     messageQueue: ConcurrentQueue[F, MessageReceived[K, M]],
     cancelToken: Deferred[F, Unit],
-    connectionFinishCallback: FinishedConnection[K] => F[Unit]
+    connectionFinishCallback: FinishedConnection[K] => F[Unit],
+    oppositeConnectionOverlap: FiniteDuration
 )(implicit tracers: NetworkTracers[F, K, M]) {
 
   private val numberOfRunningConnections = AtomicInt(0)
@@ -168,16 +171,31 @@ class ConnectionHandler[F[_]: Concurrent, K, M](
     if (conflictHappened) {
       connectionsRegister.registerIfAbsent(newConnection).flatMap {
         case Some(oldConnection) =>
-          newConnection.connectionDirection match {
-            case HandledConnection.IncomingConnection =>
-              // even though we have connection to this peer, they are calling us. One of the reason may be, that they failed and
-              // we did not notice. Lets try to replace old connection with new one.
+          if (oldConnection.age() < oppositeConnectionOverlap) {
+            // The old connection has just been created recently, yet we have a new connection already.
+            // Most likely the two nodes opened connections to each other around the same time, and if
+            // we close one of the connections connection based on direction, the node opposite will
+            // likely be doing  the same to the _other_ connection, symmetrically.
+            // Instead, let's try to establish some ordering between the two, so the same connection
+            // is chosen as the victim on both sides.
+            if (newConnection.ephemeralPort < oldConnection.ephemeralPort) {
               replaceConnection(newConnection, oldConnection)
-
-            case HandledConnection.OutgoingConnection =>
-              // for some reason we were calling while we already have connection, most probably we have received incoming
-              // connection during call. Close this new connection, and keep the old one
+            } else {
               tracers.discarded(newConnection) >> newConnection.close.as(none)
+            }
+
+          } else {
+            newConnection.connectionDirection match {
+              case HandledConnection.IncomingConnection =>
+                // Even though we have connection to this peer, they are calling us. One of the reason may be
+                // that they failed and we did not notice. Lets try to replace old connection with new one.
+                replaceConnection(newConnection, oldConnection)
+
+              case HandledConnection.OutgoingConnection =>
+                // For some reason we were calling while we already have connection, most probably we have
+                // received incoming connection during call. Close this new connection, and keep the old one.
+                tracers.discarded(newConnection) >> newConnection.close.as(none)
+            }
           }
         case None =>
           // in the meantime between detection of conflict, and processing it old connection has dropped. Register new one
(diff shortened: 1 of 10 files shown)


metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
```diff
@@ -168,16 +171,38 @@ class ConnectionHandler[F[_]: Concurrent, K, M](
...
+            // Most likely the two nodes opened connections to each other around the same time, and if
+            // we close one of the connections connection based on direction, the node opposite will
+            // likely be doing  the same to the _other_ connection, symmetrically.
+            // Instead, let's try to establish some ordering between the two, so the same connection
+            // is chosen as the victim on both sides.
+            val (newPort, oldPort) = (
+              newConnection.ephemeralAddress.getPort,
+              oldConnection.ephemeralAddress.getPort
+            )
+            val replace = newPort < oldPort || newPort == oldPort &&
```
KonradStaniec, line 185: I am just wondering if it would not be clearer to define `Order` (or `PartialOrder`) typeclass for `HandledConnection` and move this whole comparission to typeclass instance, and here only write something along the line `if (oldConnection < newConnection) replace else ...`

It would capture the intent better i.e we define the ordering between connections.

what do you think ?

KonradStaniec (commented):
Only small question, code itself looks good to me.

metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
aakoshh: I agree that it does look like a lot of logic is hidden here in private methods. I wouldn't go for a type class though, we're not trying to sort connections in general, just compare them if they are from the same peer. In that regard this class is the right place for this logic, since invoking it only makes sense if the `conflictHappened` flag has already been set.

KonradStaniec (approved):

metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala
aakoshh: So maybe an `isNewPreferredOverOldConnection` method in the connection handler that just returns a `Boolean`?

KonradStaniec: I think that would be good middle ground :+1:

aakoshh: Cool, added a `shouldReplaceConnection`.

PR #53, opened 2021-06-10
PM-3409: Interpreter client

aakoshh:
The mirror opposite of the `IntereterService` in https://github.com/input-output-hk/metronome/pull/51, the `InterpreterClient` is an RPC interface from the Service to the Interpreter.


metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/InterpreterClient.scala
```diff
@@ -0,0 +1,158 @@
...
+
+          case response: Response with FromInterpreter =>
+            rpcTracker.complete(response).flatMap {
+              case Right(ok) =>
+                tracer(InterpreterResponseIgnored(response, None)).whenA(!ok)
+              case Left(ex) =>
+                tracer(InterpreterResponseIgnored(response, Some(ex)))
+            }
+
+          case NewProposerBlockRequest(_, proposerBlock) =>
```
rtkaczyk, line 114: Could you explain how `CheckpointingService` is going to register for these notifications?

aakoshh: Sure, I was [wondering about that myself](https://github.com/input-output-hk/metronome/pull/53/files#diff-86a4b514e0cfbe03a8582cdc70f13404f7b07bffc66d69652d064e7aa874823eR135-R136). 

Conceptually the `ServiceRPC` would be implemented by the `CheckpointingService`, and it will need a reference to an `InterpreterRPC` to send commands and queries. That seems circular because to create the `IntepreterRPC` you need to pass a `ServcieRPC` to `InterpreterClient.apply`. 

I thought it would be resolved by sharing common references to things. Ignoring all other details here:

```scala

class CheckpointingService[F[_]: Sync](
  mempoolRef: Ref[F, Mempool],
  interpreterRpc: InterpreterRPC[F],
  ...
) extends ApplicationService[F, CheckpointingAgreement]

object CheckpointingService {
  def apply[F[_]: Concurrent](
    localConnectionManager: InterpreterConnection[F],
  ): Resource[F, ApplicationService[F, CheckpointingAgreement]] = 
    for {
      mempoolRef <- Resource.liftF(Ref[F].of(Mempool.empty))
      serviceRpc = new ServiceRpc[F] {
        override def newProposerBlock(proposerBlock: Transaction.ProposerBlock) = 
           mempoolRef.update(_ :+ proposerBlock)
      }
      interpreterRpc <- InterpeterClient(localConnectionManager, serviceRpc)
      service = new CheckpointingService(mempoolRef, interpreterRpc)
    } yield service
}
```
If something needed responses, it would be more complicated, involving a `ConcurrentQueue` and `Deferred`s that the `service` would have to process.

rtkaczyk: > ServiceRPC [...] will need a reference to an InterpreterRPC to send commands and queries. That seems circular

Yeah, I was confused by that. My understanding is that the command/query you mention is `CreateBlockBodyRequest` which is sent subsequently to receiving `NewCheckpointCandidateRequest`.

Firstly, when is `CreateBlockBodyRequest` actually sent? Is it directly in response to `NewCheckpointCandidateRequest`, provided that we perform some sort of leader check? If so I would imagine we would register a callback with `ServiceRPC` so that it doesn't need a reference to `InterpreterRPC`.

Secondly, why does the processing of incoming notifications has to happen within `InterpreterClient`? If it was in the `ServiceRPC` impl, then the inverse reference would also become unnecessary.

aakoshh: Sorry, I wanted that to read `CheckpointingService [...] will need a reference to an InterpreterRPC`. The `ServiceRPC` is a handcover mechanism that the `InterpreterClient` needs to handle the stuff coming from the Interpreter over the connection. 

> when is CreateBlockBodyRequest actually sent?

`CreateBlockBodyRequest` is sent in `CheckpointingService.createBlock`, which, as you say, will only be called if the node is the leader, so you don't need to perform more checks. 

It's not in response to `NewCheckpointCandidateRequest`. The latter we could rename to `NewCheckpointCandidateNotification`, but I decided not to introduce new naming; in any case it's a [Request with NoResponse](https://github.com/input-output-hk/metronome/blob/develop/metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/messages/InterpreterMessage.scala#L55). The purpose of this notification is just a possible optimisation: we can use it to _only_ send the create body request if there is even the potential that something will come out of it; otherwise we can just create an empty block without pestering the interpreter, spending some nominal amount of time to not churn out blocks at light speed. 

> If so I would imagine we would register a callback with ServiceRPC so that it doesn't need a reference to InterpreterRPC.

I thought handling this particular call would be as simple as setting a `Ref[F, Boolean]` to `true`, so that next time the node is leading it checks in. This is what I did in the tests.

If it had to send a direct response, then it the `ServiceRPC` handler method could create a `Deferred` and put the request in a queue, which the `CheckpointingService` can read and complete with some answer.

> why does the processing of incoming notifications has to happen within InterpreterClient

The `InterpreterClient` just takes messages from the queue and hands them over to the `ServiceRPC`, so effectively message handling _is_ happening in the `ServiceRPC` implementation. The `ServiceRPC` can return an answer right there in its methods if it can, like it happens on the [opposite side](https://github.com/input-output-hk/metronome/pull/51/files#diff-78ed33bb8a82a76af09602e757c346500202fde7c5729a2f5a323a8aa47a22f6R57). That's not the reason why the `CheckpointingService` needs an `InterpreterRPC` reference. 

The reason for that is so that the `CheckpointingService` can send messages to the Interpreter any time it wants, for example to validate a block.

aakoshh: The way I think about it is this:
* As the `IntepreterClient` I can send `Request with FromSerivce` to the `LocalConnectionManager` and read `Response with FromInterpreter`, therefore I can implement `InterpreterRPC`.
* As the `IntepreterClient` I can read `Request with FromInterpreter` from the `LocalConnectionManager` and send `Response with FromService`, therefore I need a `ServiceRPC` implementation to hand over the processing to.
* As the `CheckpointingService` I need an `InterpreterRPC` implementation, so that I can send requests to the Interpreter and receive responses whenever I have to.
* As the `CheckpointingService` I want to be notified about the `Request with FromInterpreter with NoResponse` messages coming from the Interpreter; in principle I could implement `ServiceRPC`.
* If as `CheckpointingService` gets a reference to the return value of `IntepreterClient` then it cannot also be passed to it, so the `ServiceRPC` implementation needs to be separate, and created first.
* The `ServiceRPC` implementation therefore needs to have access to some shared memory with the `CheckpointingService.`

rtkaczyk: > I thought handling this particular call would be as simple as setting a Ref[F, Boolean] to true, so that next time the node is leading it checks in. This is what I did in the tests.

Cool, so all we need from `ServiceRPC` is to do state accumulation. Should be straight-forward :+1: 


> The InterpreterClient just takes messages from the queue and hands them over to the ServiceRPC, so effectively message handling is happening in the ServiceRPC implementation

Sorry, that's what I meant. Why are we introducing that delegation? Why couldn't the `ServiceRPC` be the one that reads incoming messages? Is there something about `LocalConnectionManager` that I'm missing?

aakoshh: So you are asking why `InterpreterClient` exists, with the alternative being that the `CheckpointingService` can just get the `LocalConnectionManager` and send/handle all the messages on its own. 

You are right, this could be done. The result would be something like [SyncService.processNetworkMessages](https://github.com/input-output-hk/metronome/blob/develop/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala#L132). 

The reason I thought we should delegate is that the `CheckpointService` will need to have a remote and a local message handling interface as well (remote to sync state with other federation members, like the `RobotService`) and it could grow massive. I thought we can formalise the local interaction, with all the timeouts, RPC message ID pairing, into a class that can be tested on its own, to keep the complexity of the `CheckpointingService` under control. 

The `LocalConnectionManager` itself is not special, just a restricted version of the `RemoteConnectionManager`. I really just wanted to encapsulate things to make it easier to understand and maintain, and hopefully to code the `CheckpointingService` as well as for example in `validateBlock` you'd just have to call a method that looks like regular RPC.

aakoshh: Another way to interpret your question in a more narrow sense is why specifically the `ServiceRPC` is not reading the messages from the `LocalConnectionManager`, leaving everything else as it is now, which is that the `IntepreterRPC` implementation would still live in the `InterpreterClient`.

This could also be done I suppose. The difficulty is that, there's just 1 message queue for both requests and responses coming from the Interpreter. The `InterpreterRPC` needs to get the responses, the `ServiceRPC` needs to get the requests. We could separate the queue with a mechanism like the `Network.splitter` method. 

I didn't think about that. I think the `InterpreterClient` would still pack a lot of boilerplate with timeout handling and request ID matching, but in this case where the `ServiceRPC` only deals with notifications, it could be a simpler model to grasp.

rtkaczyk: > So you are asking why InterpreterClient exists

No, I'm not asking that.

> why specifically the ServiceRPC is not reading the messages from the LocalConnectionManager

Yes, that's what I'm asking. My understanding is that `InterpreterRPC` handles calls initiated from the Checkpointing Service to the Interpreter, and vice-versa for the `ServiceRPC`. It seems to me those both those RPCs could just share a common reference to `LocalConnectionManager` and they wouldn't need to be coupled otherwise.

Now, I haven't delved into the implementation or `LocalConnectionManager` or any underlying networking code, but it seems to me that since we can _concurrently_:
* `sendMessage` and match it with a response
* read `incomingMessages` directly

then nothing should prohibit us from putting those operations into separate classes. But you are saying it can't be done without some additional coding to split the messages, is that right? If so, then I wouldn't think it's worth it, but could you explain that a bit more?

Alternatively, I don't suppose it would make sense to maintain separate connections for `ServiceRPC` and `InterpreterRPC`, would it?

aakoshh: > But you are saying it can't be done without some additional coding to split the messages, is that right? If so, [...] could you explain that a bit more?

Sure, it's very simple: the way Konrad implemented the [ConnectionHandler.incomingMessages](https://github.com/input-output-hk/metronome/blob/develop/metronome/networking/src/io/iohk/metronome/networking/ConnectionHandler.scala#L125-L126) is that it repeatedly pops messages off a single `messageQueue`, so if we tried to use multiple instances of the `Iterant` they'd be stealing messages from each other. 

This is different from how the underlying Scalanet channels work, they have a separate message queue per connection, but it's lower level API where you have to handle incoming connections. One of the main goals of the `RemoteConnectionManager` was to make sure that there is at most one connection between federation members, so if they both open connections to each other at the same time, we keep only one. The other was to keep the federation fully to connected. It ended up completely abstracting away the connections and merging everything into a single message stream, regardless of connections coming and going. 

I use `Network.splitter` construct in to multiplex messages and route them to the right handler class. It helps with the compiler as well as it allows the different components to consume messages at their own pace.

In this case that thought didn't kick in for me because I thought that basically we only have `CheckpointingService` that wants to talk to the Interpreter.  The messages are also part of a single ADT. To be fair, if the `CheckpointingService` would have to send responses, then as I described earlier we'd need a message queue to hand over messages to it from the anonymous `ServiceRPC`, at which point it would be almost the same as the splitter (just 2 queues, not 3), so it's _almost_ worth it :-)

> Alternatively, I don't suppose it would make sense to maintain separate connections for ServiceRPC and InterpreterRPC, would it?

I remember the Cardano team saying that one should avoid queueing, so they use STM instead, yet here I am with all these queues :-/ But they do multiplex a lot of protocols over a single TCP connection, and I thought that's an elegant design. Configuring multiple ports for every little thing we want to exchange messages for would be difficult to maintain I think, easy to missconfigure some ports and end up with cross talk.

rtkaczyk: Thanks, that all makes sense.
Somehow I missed the fact the `sendMessage` relies on `RPCTracker` to complete the responses. I thought it was using some under-the-hood magic and that matching the responses was done independently from reading `incomingMessages`. Hence my question. 

Here's another one: wouldn't it make sense to have `private class ClientImpl[...](...) extends InterpreterRPC with ServiceRPC`?

aakoshh: Haha, no, this is a distinctively no-magic solution :sweat_smile: I added `sendRequest` to `SyncService` to capture the mechanisms of a request-response cycle, and it could probably be abstracted away into some generic component that could be reused between `SyncService`, `CheckpointingService` and the `InterpeterClient`; it's an accident that the Interpreter SDK didn't need it.

But I have an idea how we could add magic :mage_man:  Here's a [library](http://alcestes.github.io/lchannels/) for using [Session Types](http://www.dcs.gla.ac.uk/research/betty/summerschool2016.behavioural-types.eu/programme/DardhaIntroBST.pdf/at_download/file.pdf) with Scala, with a short example on the project page. I thought we could implement Session Types with a Free monad, with a few low level primitives like `Receive[T]` and `Send[T]` on the network, and an `Eval[T]` for evaluating user-land functions that return `T`; this we could pass to an interpreter which hides the magic of matching requests and responses somehow, under combinators like `Ask[Req, Res]`. Then we wouldn't have stuff like the `RPCTracker` appear in our classes, nor the `Network` with explicit message sending and handling, just a big `for` comprehension to describe our protocol. 

Anyway, back to Earth.

Would `ClientImpl extends InterpreterRPC with ServiceRPC` make sense? 

Since this is a private class, we'd rather have to see first what (roughly) `InterpeterClient.apply(s: ServiceRPC): InterpreterRPC` would look like in this case. I don't think returning a `ServiceRPC` from that method would make sense because we wouldn't want to call it from anywhere within our application; we want those methods to be called when the client receives requests, so I think it makes sense that `ServiceRPC` is an input there. 

But would it make sense for `ClientImpl` to extend it? I reckon the only way it could do so is if instead of a `ServiceRPC` input it would receive all dependencies that it needs to perform the tasks we want it to do, e.g. the actual mempool in this case. But then it wouldn't _really_ have to implement it, just handle the messages directly; it might provide some clarity to implement the interface and pass to handlers internally, but nobody else would call it. 

I think doing it this would be introduce extra coupling though and break the symmetry with the Interpreter SDK; we may have similar conversations with the ETC team soon :smile: 

It seems to me that the separation of concerns that queue splitting could bring could be a better de-mystifier.

rtkaczyk: > Anyway, back to Earth

:D really cool!

> But then it wouldn't really have to implement it, just handle the messages directly; it might provide some clarity to implement the interface and pass to handlers internally, but nobody else would call it.

:+1:

> It seems to me that the separation of concerns that queue splitting could bring could be a better de-mystifier.

I agree, but it seems a lot of work for little gain.

Thanks for your explanations

metronome/checkpointing/service/test/src/io/iohk/metronome/checkpointing/service/InterpreterClientSpec.scala
```diff
@@ -0,0 +1,333 @@
...
+    implicitly[Arbitrary[T]].arbitrary.sample.get
+
+  def test(fixture: Fixture): Assertion = {
+    implicit val scheduler = TestScheduler()
+
+    val fut =
+      fixture.resources.use(fixture.test).delayExecution(0.second).runToFuture
+
+    // These tests shouldn't take too long, so just simulate some generous amount of time.
+    scheduler.tick(1.minute)
```
rtkaczyk, line 46: I'm curious why this (and `delayExecution`) is needed

aakoshh: Excellent question. I'm sorry, I left a [comment](https://github.com/input-output-hk/metronome/pull/44/files#diff-31f9fae33dbc8e276e1061b50f1d73e06c1f904ab986f8948509de386f4fda76R33) in the `RobotIntegrationSpec`. 

When I first used the `TestScheduler` I noticed that some of the `Task`s I schedule get executed when I call `tick()` but others run earlier. The reason was that if a task has no delay, it will be just executed right there, so the `TestScheduler` can work through everything a delayed task triggered until it hits something that is scheduled for the future. 

Not sure why it was a problem in those tests I was working on, but I thought it would be easier to understand things if I set up the test and know that it will only do something when I pull the trigger. Adding even a 0 delay means nothing happens until this line.

Here's a demonstration:

```console
$ mill -i metronome[2.13.4].core.console
[31/31] metronome[2.13.4].core.console 
Welcome to Scala 2.13.4 (OpenJDK 64-Bit Server VM, Java 11.0.11).
Type in expressions for evaluation. Or try :help.

scala> import monix.execution.schedulers.TestScheduler
import monix.execution.schedulers.TestScheduler

scala> import monix.eval.Task
import monix.eval.Task

scala> val t = Task(println("Hello"))
val t: monix.eval.Task[Unit] = Task.Eval$1605702850

scala> implicit val scheduler = TestScheduler()
val scheduler: monix.execution.schedulers.TestScheduler = monix.execution.schedulers.TestScheduler@215c6ec0

scala> val f = t.runToFuture
Hello
val f: monix.execution.CancelableFuture[Unit] = monix.execution.CancelableFuture$Pure@3ba8ef4e

scala> import scala.concurrent.duration._
import scala.concurrent.duration._

scala> val t0 = Task(println("World")).delayExecution(Duration.Zero)
val t0: monix.eval.Task[Unit] = Task.FlatMap$719355325

scala> val f0 = t0.runToFuture
val f0: monix.execution.CancelableFuture[Unit] = Async(Future(<not completed>),monix.eval.internal.TaskConnection$Impl$$anon$1@3100b5cf,null)

scala> scheduler.tick()
World
```

As to why `scheduler.tick(1.minute)` is needed: there are timeouts involved in some of the tests, for example [this one](https://github.com/input-output-hk/metronome/pull/53/files/938f5b56ac79bce6b7ffb8e180411a38cb476e55#diff-080bb23b02bac10ccb5fd2b77c71f79c6f5d8fcf52c26543902b0fb9ecc90c97R98). These tests all use mocks, not a real network (unlike the ones in https://github.com/input-output-hk/metronome/pull/51 which start an actual Scalanet peer groups), so we don't actually have to wait out the 5 second timeout, or set them to reasonably low values; we can use the `TestScheduler` to simulate the passage of time. 

In the `RobotIntegrationSpec` I added a [duration property](https://github.com/input-output-hk/metronome/pull/44/files#diff-31f9fae33dbc8e276e1061b50f1d73e06c1f904ab986f8948509de386f4fda76R175-R178) to the `Fixture` to say how long it's going to run for, but here they are all just "some time with a potential timeout" so 1 minute should be plenty. 

I don't want to just keep ticking until there are no more tasks because there could be something on a loop, so this gives a timeout for the test as well. It won't actually do anything if there are no tasks.

aakoshh: It's nice in that when I call `enqueueMessage` it will add a message to a `ConcurrentQueue` which is consumed on a different `Fiber`, so I don't know exactly when I can expect the effects to be observable. I know that the execution is _almost_ instantaneous, but not quite.

I could maybe add some `Deferred` somewhere that I can wait upon, but it could be awkward to do. Or I can call `Task.sleep` in the test, but how long should I sleep for? What works locally always turns out to be too tight on CI.

With the `TestScheduler`, the non-delayed tasks will be executed first, so I can add something like this `awaitProcessing` with a sleep time that I can trust is surely enough for something that is processed immediately, and then all it will do is jump the clock by 5 seconds and carry on with the rest of my test, doing the assertions, without actually wasting any time.

rtkaczyk: Interesting, thanks for the explanation

rtkaczyk (approved):
As usual, thank you very much for the time taken to help me understand what's going on

PR #54, opened 2021-06-14
PM-3110: Create checkpoint certificate

rtkaczyk:


metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -51,7 +57,7 @@ class CheckpointingService[F[_]: Sync: Applicative, N] private (
   override def executeBlock(
       block: Block,
       commitQC: QuorumCertificate[CheckpointingAgreement],
-      commitPath: NonEmptyList[Block.Hash]
+      commitPath: NonEmptyList[Block.Hash] //TODO: remove?
```
aakoshh, line 60: I added this in case an application would need to construct a proof by retrieving block headers along the way, or something similar. Not every use case must use it, but I think it makes sense to provide the link between the `block` and the `commitQC` (which is also use case specific).

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -139,15 +147,47 @@ class CheckpointingService[F[_]: Sync: Applicative, N] private (
...
+  //TODO: could be optimised for no checkpoint case
+  //TODO: avoid .get
+  private def contructCheckpointProof(
+      block: Block,
+      ledger: Ledger
+  ): Option[CheckpointWithProof] = {
+    val tree = MerkleTree.build(
+      block.body.transactions.map(tx => MerkleTree.Hash(tx.hash))
+    )
+    val checkpoint     = ledger.maybeLastCheckpoint.get
```
aakoshh, line 186: You can avoid the `.get` by just finding the last checkpoint in the block body. Arguably there's nothing to prove if there was no checkpoint added in the block, so constructing a Merkle tree and then trying to find the checkpoint in the ledger that could be from a previous state is more work than finding the the checkpoint and constructing a proof from it. Shouldn't even need the `ledger` for this, unless as a sanity check that the last checkpoint in the block matches the ledger.

Also, there may not be a `maybeLastCheckpoint` in the ledger yet, in case we just added the first proposer block, or we're just committing an empty block after genesis. We can't start with a checkpoint because we have no idea what's on the PoW side.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -139,15 +147,47 @@ class CheckpointingService[F[_]: Sync: Applicative, N] private (
...
+    def traverseCommitPath(
+        ledgers: LedgerTree,
+        from: Block.Hash,
+        path: NonEmptyList[Block.Header]
+    ): Option[CheckpointCertificate] =
+      ledgers.get(from) match {
+        case Some((_, _, Some(cp))) =>
+          CheckpointCertificate(path, cp.checkpoint, cp.proof, commitQC).some
+        case Some((_, h, None)) =>
+          traverseCommitPath(ledgers, h.parentHash, path :+ h)
```
aakoshh, line 168: Maybe it would make sense to reverse the order in `CheckpointCertificate` so `head` is where the checkpoint is and `last` is what the commit is about, so blocks are in natural order?

I thought the order from commit to checkpoint would facilitate checking, but don't have a strong opinion. 

I would be nice to prepend instead of append though.

aakoshh (commented):
The state accumulation looks good :+1:

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: What if we had a special value for genesis so that we can avoid the `Option`, like:

```
val GenesisCheckpointCandidate = CheckpointCandidate(BitVector.empty)
```
?

I realise that building the proof is not optimal, as more often it will be the case that we don't have a checkpoint in the block. It was just nicer to code this way, but I will update this.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: I think I would prefer the natural order. I don't think the order should really affect the complexity of checking the parent links.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: If we use `BitVector.empty` to indicate that this is a genesis, then I guess the problem is that 
1. We assume that the checkpointing interpreter doesn't itself want to use `BitVector.empty`, whereas so far we said we don't try to guess what's in a checkpoint.
2. We exchange a compiler supported condition of having no checkpoint so far with another kind of `.isEmpty`, but one we can forget. For example here you would get rid of a `.get`, but we would be sending the `Ledger` in all validation and block creation requests, and want the interpreter implementer be aware while checking for example that we are extending the previous checkpoint in the ledger that it may not be a checkpoint after all. At least we probably can't produce an invalid certificate because that would require the body to have one like this as well, which would hopefully fail to validate.

I thought turning `Ledger` into an ADT would be an overkill, but that would be another option, that we are either before we got our first checkpoint, or after.

aakoshh:
As we discussed it would be good to have tests for the creation of a checkpoint certificate, and it could live outside the service, for example in the `CheckpointCertificate` companion, given all the input values to construct the proof. 

This would also give you the opportunity to add a validation function as well, which can be used by the interpreter implementations, since they will use the `checkpointing.models` package as a library.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: No longer using `.get` so that part's resolved.

But I'm still thinking we could use a `GenesisCheckpointCandidate`. Its value could be agreed upon via genesis configuration. This is similar to how we define a genesis block. We could also have `parentHash: Option[Hash]`, but that would be cumbersome in the long run, so we specify some artificial value. The PoW interpreter could propose such a value and we'd include it in the definition of the genesis block in metronome.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -135,18 +149,46 @@ class CheckpointingService[F[_]: Sync: Applicative, N](
...
+
+    def traverseCommitPath(
+        ledgers: LedgerTree,
+        from: Block.Hash,
+        path: List[Block.Header]
+    ): Option[CheckpointCertificate] =
+      ledgers.get(from) match {
+        case Some((_, h, Some(cp))) =>
+          val headers = NonEmptyList.of(h, path: _*)
+          CheckpointCertificate(headers, cp.checkpoint, cp.proof, commitQC).some
```
rtkaczyk, line 168: Regarding:

> As we discussed it would be good to have tests for the creation of a checkpoint certificate, and it could live outside the service, for example in the CheckpointCertificate companion, given all the input values to construct the proof.

(possibly a longer discussion, so moving it here)

It's easy to move the proof construction to the companion object. But we still need to traverse the headers to construct the certificate. We could make this construction dependent on a function that returns an optional parent header, but we'd also have to verify the merkle proof on each step as final traversal condition.

Do you think we should do that?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -60,14 +68,12 @@ class CheckpointingService[F[_]: Sync: Applicative, N](
     updateLedger(block).flatMap {
       case Some(ledger) =>
         if (block.hash != commitQC.blockHash)
-          saveState(block.header, ledger).as(true)
+          false.pure[F]
```
rtkaczyk, line 71: As you observed in #45 if we were to save this block as the last executed we could potentially lose the checkpoint information from preceding blocks. Alternatively, to facilitate persisting execution of each (non-committed) block, we could modify the ledger to also contain the checkpoint proof, or at least the containing header's hash.

WDYT?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -60,14 +68,12 @@ class CheckpointingService[F[_]: Sync: Applicative, N](
...
         else
           for {
-            //TODO: PM-3110:
-            // chkpOpt <- constructCheckpoint(ledger, commitQC)
-            _ <- saveState(block.header, ledger)
-            //TODO: PM-3110:
-            // _ <- chkpOpt.map(pushCheckpoint).getOrElse(().pure[F])
+            chkpOpt <- constructCheckpointCert(commitQC)
+            _       <- saveState(block.header, ledger)
+            -       <- chkpOpt.map(pushCheckpoint).getOrElse(().pure[F])
```
rtkaczyk, line 76: If the app crashes at this point we will push the checkpoint on restart. Maybe we should have another marker for whether the push was completed? 🤔 

To be solved in PM-3137 anyway

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -155,12 +197,19 @@ class CheckpointingService[F[_]: Sync: Applicative, N](
 }
 
 object CheckpointingService {
-  type LedgerTree = Map[Block.Hash, (Ledger, Block.Header)]
+  type LedgerTree =
+    Map[Block.Hash, (Ledger, Block.Header, Option[CheckpointWithProof])]
```
rtkaczyk, line 201: Alternatively we could just save the last proof along with the block hash in a separate Ref 🤔

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: I'm not against putting the identity of the PoW genesis block into the starting ledger as the first checkpoint, but I'd think about extending the RPC interface where we can query the interpreter for this information upon startup, rather than try to manage it manually in configuration. 

@dzajkowski already asked if the checkpoints can have the block height as well as the identity in them, and sure enough, the interpreters can encode any content into the binary values inside the checkpoint. A height of 0 could indicate that this is the genesis candidate, and the interpreter can use this information to validate the block in a different way.

Because of that, I don't know why we would need a separate data type like a `GenesisCheckpointCandidate`. 

I'm also not sure this brings much to the table in terms of extra knowledge that the interpreter wouldn't have if we don't add it: in order to trust this value, it must come from the interpreter, and inside the interpreter it can always be used where we had `None` so far. The only spot where it provides something new that you can immediately tell that you the other federation members are on a different chain, if the hashes don't match, so you cannot extend it with a new checkpoint. But this can only happen at the beginning, after that they move on, and you just stand there not knowing the checkpoints they are talking about at higher levels.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: I wouldn't start passing functions for optional parent headers, I was more thinking about a pure function.

TBH the way you do it is so different from what I had in mind originally that it's not worth moving around. What I was thinking of was to have a buffer where you accumulate the inputs from the last block you encountered that had a checkpoint candidate to the block which has a commit, and then you only produce a proof at the end, so it would be something like `case class CheckpointCertificateInputs(lastBlockWithCheckpointCandidate: Block, blocksSinceLastCheckpointCandidate: List[Block.Header)`. 

Then it would be easy to just pass this block and the headers, along with the commit certificate, to a prover, it wouldn't have to retrieve anything else.

But because you are constructing a checkpoint proof for every block you see, and later go back to find the last one that had a non-empty proof in it, the logic is meshed into the whole service, not as clean to remove. So I wouldn't bother with moving just an iteration logic that has to retrieve headers until it finds something interesting.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: See this [comment](https://github.com/input-output-hk/metronome/pull/54/files/1a3f15eb976f58996f5249b52991d8585ac81380#r661533556), I reckon you would also need the headers along the way.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: I think it's fine this way.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: Right, because if we crash before returning from `executeBlock` then the `BlockExecutor` will not remember that this block was indeed executed. However it will wait for the next commit certificate to arrive, at which point there may be newer candidates to push. But if there aren't, you get a longer proof for the same candidate. 

This longer proof thing can happen to other nodes as well that somehow miss a Q.C. and construct a proof based on the next block. It's the malleability issue in a different disguise, and the interpreter has to be able to handle it.

So I wouldn't bother with the completion, especially since pushing the message doesn't mean it was received, it's very best effort.

aakoshh (approved):
Looks correct! 

As discussed, the `CheckpointCertificate` could capture more of the proof generation from the raw inputs, so that it doesn't have to be replicated in `expectedCheckpointCert` for example, but I leave it up to you which approach you prefer.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Well, it would be difficult to maintain an accumulated list. Unless we want to differentiate `updateLedger` between validation and execution we need to maintain a tree, as we already do. In either case the logic of building the path part of the proof would reside mostly in `CheckpointingService`. So really the logic of the potential smart constructor boils down to constructing the merkle proof.

I agree it would be nice to have something like `CheckpointCertificate.construct(block, headers, commitQC)`. The merkle proof would be constructed in that method. However, that would require to keep potentially multiple full blocks in memory (different branches) or looking up the block in storage. I wanted to avoid either.

I'd still put the merkle proof construction in a common place like `CheckpointCertificate` so that I can reuse it in tests.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Forget it. We need to maintain a tree and we don't really know which one is the _last one_ until we see a commit QC.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: `GenesisCheckpointCandidate` - I don't know why I wrote it like that. I meant just the concept of having a genesis value, it certainly doesn't need to be expressed by another type. And I like the idea of querying the interpreter for it, rather than relying on config :+1:

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: Exactly, I thought this list accumulation would only happen in execution, not validation, it only makes sense there. Only the headers would have been kept in memory and the body of the last block that had a checkpoint. 

But I don't want to upset your scheme. Moving the merkle proof construction to a reusable place is a good step.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
aakoshh: The executed blocks are always the last ones, they are not going to be un-executed. Only validation is tentative. Execution is triggered by a commit of a descendant-or-self.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Ok, you're right. Duh, I can move the accumulation outside of `updateLedger`. I will do that

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Ok, you're right. Duh, I can move the accumulation outside of updateLedger. I will do that

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Done, the checkpoint-relevant accumulation only happens in `executeBlock` now, and certificate construction code has been moved to the companion

aakoshh (approved):
Looks ace! It was really worth doing the refactorings to pull out the common bits, makes the service and the tests less busy.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -68,14 +77,12 @@ class CheckpointingService[F[_]: Sync, N](
     projectLedger(block).flatMap {
       case Some(ledger) =>
         if (block.hash != commitQC.blockHash)
-          saveLedger(block.header, ledger).as(true)
+          updateCheckpointData(block).as(false)
```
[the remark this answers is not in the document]
rtkaczyk: I kinda wanted to hide the `Option` handling, which makes the code a bit awkward at times. See if you like it better now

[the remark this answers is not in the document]
aakoshh: Cool! You always manage to find more tools in the shed, like this `cataF`. 
Having the update happen unconditionally is I think more intuitive, thank you :+1:

aakoshh:
@rtkaczyk I created another story https://jira.iohk.io/browse/PM-3562 to do checkpoint certificate validation, so the Interpreter implementors don't have to do it themselves.

PR #55, opened 2021-06-15
PM-3418: Add height to the Block type class. Validate it in SyncService.

aakoshh:
https://github.com/input-output-hk/metronome/pull/45 added a `height` to the checkpointing block type to avoid the possibility of traversing back to the root of the tree. I thought this would be a good idea in general as well, since `BlockStorage.getPathFromAncestor` could also suffer from this, and it's easy enough to support this value in any block tree.

The PR adds `height` to the `hotstuff.consensus.basic.Block` type class, so everything needs to implement it.

I also noticed while looking for a place where this can be validated that I never called `Block.isValid` on the proposed block, only its parent. This has been added as well. 

This means we no longer need the opt-in solution of `maxDistance` in `BlockStorage.getPathFromAncestor`, which was added in https://github.com/input-output-hk/metronome/pull/39

I also changed the underlying child-to-parent collection to store block meta-data instead, so we can access the parent and the height without reading the full block.


metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
```diff
@@ -253,6 +254,25 @@ class SyncService[F[_]: Concurrent: ContextShift, N, A <: Agreement: Block](
...
+    for {
+      parent <- OptionT {
+        storeRunner.runReadOnly {
+          blockStorage.get(Block[A].parentBlockHash(block))
+        }
+      }
+      // Reject the block straight away if it doesn't pass basic validation.
+      isFormallyValid =
+        Block[A].height(block) == Block[A].height(parent) + 1 &&
+          Block[A].isValid(block)
```
rtkaczyk, line 268: There are 3 kinds of validations in this function:
- _consensus-specific syntactic validation_: height
- _application-specific syntactic validation_: content root
- _application-specific content validation_: interpreter-assisted

I'm worried that currently these look a bit like they're tied with a string. Perhaps we could capture these abstractions with better naming, but I realise that them being such simple checks this may look like an overkill.

Not a blocker for this PR, but I thought I'd point this out.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
```diff
@@ -253,6 +254,25 @@ class SyncService[F[_]: Concurrent: ContextShift, N, A <: Agreement: Block](
         .void
   }
 
+  /** Validate the prepared block after the parent has been downloaded. */
+  private def validateBlock(block: A#Block): OptionT[F, Boolean] =
```
rtkaczyk, line 258: I realise this makes the code slightly more concise, but I tend to think of monad transformers as tools to be used in-place rather than passed around.

Not a blocker, but I'm curious about your opinion.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockStorage.scala
```diff
@@ -14,22 +14,25 @@ import scala.collection.immutable.Queue
   */
 class BlockStorage[N, A <: Agreement: Block](
     blockColl: KVCollection[N, A#Hash, A#Block],
-    childToParentColl: KVCollection[N, A#Hash, A#Hash],
+    blockMetaColl: KVCollection[N, A#Hash, BlockStorage.BlockMeta[A]],
```
rtkaczyk, line 17: 👍

rtkaczyk (approved):
This will be useful for #45  👍

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Thanks, I also would have liked to see it broken up a bit, but didn't have any good ideas. Maybe I can move the the height check into an `isParentOf` method into the `Block`, so it doesn't need to be repeated. 

What exactly do you mean by "tied with a string"?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: I only did it this way because I didn't like to do this:
```scala
def validate = 
  (for {
    a <- OptionT(foo)
    b <- OptionT(bar)
  } yield a+b).value
```
or this:
```scala
def validate = { 
  val howDoINameThis = for {
    a <- OptionT(foo)
    b <- OptionT(bar)
  } yield a+b

  howDoINameThis.value
}
```

I also think that if there are too many `OptionT` values in the expression it's actually better to pass them around, and only take the `.value` at the very end, rather than having to keep rewrapping them. The goal IMO is to make things readable, to make the extra syntax dealing with `Option` disappear. 

I remember that it was really confusing to see all the `lift` and `EitherT` everywhere in a massive `for` expression, when I first saw them being applied, it obscured the logic. 

Another reason I thought this could make a difference is that both methods are just called `validateBlock`, so I thought it lowers the chance of mixing up `appService.validateBlock` with this if they have a different return type.

I don't have a strong opinion in this case, it just felt less awkward.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Tried something, not sure it makes it much better.

aakoshh: And another variant, maybe this one's good, there's no combination.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala
aakoshh: Removed `OptionT` in https://github.com/input-output-hk/metronome/pull/55/commits/57d08834ba4c9d041036f1d468fd4d085dc997e3

PR #56, opened 2021-06-16
PM-3419: Generalise BlockStorage into KVTree

aakoshh:
Extracted all the generic logic from `BlockStorage` into a `KVTree` component that the former now inherits. Testing is still through `BlockStorageProps` - I would have to test with some test data anyway, it's just in a different module.

This could facilitate storing things that follow the block tree structure (such as the tentative ledger) and reuse methods like pruning.

Builds on https://github.com/input-output-hk/metronome/pull/55 because the `height` is needed to avoid unnecessary traversals for ancestors now.


lemastero (approved):
LGTM 👍

PR #58, opened 2021-07-12
publish config module

biandratti:
Adding the option to publish config module

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -62,7 +62,8 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
         Developer("aakoshh", "Akosh Farkash", "https://github.com/aakoshh"),
         Developer("lemastero","Piotr Paradzinski","https://github.com/lemastero"),
         Developer("KonradStaniec","Konrad Staniec","https://github.com/KonradStaniec"),
-        Developer("rtkaczyk", "Radek Tkaczyk", "https://github.com/rtkaczyk")
+        Developer("rtkaczyk", "Radek Tkaczyk", "https://github.com/rtkaczyk"),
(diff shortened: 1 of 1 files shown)


build.sc
```diff
@@ -62,7 +62,8 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
         Developer("aakoshh", "Akosh Farkash", "https://github.com/aakoshh"),
         Developer("lemastero","Piotr Paradzinski","https://github.com/lemastero"),
         Developer("KonradStaniec","Konrad Staniec","https://github.com/KonradStaniec"),
-        Developer("rtkaczyk", "Radek Tkaczyk", "https://github.com/rtkaczyk")
+        Developer("rtkaczyk", "Radek Tkaczyk", "https://github.com/rtkaczyk"),
+        Developer("biandratti", "Maxi Biandratti", "https://github.com/biandratti")
```
aakoshh, line 66: :wave:  :partying_face:

aakoshh (approved):
One thing to keep in mind is that we'll have to start bumping the version of metronome after this snapshot, so that Nix doesn't complain about the changing hash of `0.1.0-SNAPSHOT`. 

Maybe it's time to merge into `master` and release `0.1.0` so you don't have to reference the `SNAPSHOT` version.

biandratti:
> One thing to keep in mind is that we'll have to start bumping the version of metronome after this snapshot, so that Nix doesn't complain about the changing hash of `0.1.0-SNAPSHOT`.
> 
> Maybe it's time to merge into `master` and release `0.1.0` so you don't have to reference the `SNAPSHOT` version.

great, I'll do it

PR #59, opened 2021-07-13
[PM-3109/3131/3132/3137] CheckpointingService calls to Interpreter

rtkaczyk:


metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -141,16 +154,25 @@ class CheckpointingService[F[_]: Sync, N](
...
-      block: Block
+  private def updateLedgerTree(
+      ledger: Ledger,
+      block: Block,
+      //TODO: I don't like this
+      precomputed: Boolean = false
   ): F[Ledger] = {
-    val newLedger = oldLedger.update(block.body.transactions)
+    val newLedger =
+      if (precomputed) ledger else ledger.update(block.body.transactions)
```
aakoshh, line 175: I would not have guessed that `precomputed` means the ledger is for the block, otherwise it's for the parent. Maybe a separate method, or an `Either` (left=parent, right=child ledger)?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -30,19 +32,38 @@ import scala.annotation.tailrec
...
+        else
+          interpreterClient.createBlockBody(oldLedger, mempool.proposerBlocks)
+      }
+
+      newLedger = oldLedger.update(newBody.transactions)
+      newBlock  = Block.make(parent.header, newLedger.hash, newBody)
+
+      _ <- OptionT.liftF(
+        updateLedgerTree(newLedger, newBlock, precomputed = true)
+      )
```
aakoshh, line 64: Once the block is created, you can set `hasNewCheckpointCandidate` to false.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/Mempool.scala
```diff
@@ -0,0 +1,15 @@
...
+case class Mempool(
+    proposerBlocks: IndexedSeq[ProposerBlock],
+    hasNewCheckpointCandidate: Boolean
+) {
+  def isEmpty: Boolean =
+    proposerBlocks.isEmpty && !hasNewCheckpointCandidate
+}
+
+object Mempool {
+  val empty: Mempool = Mempool(Vector.empty, false)
```
aakoshh, line 14: It might be safer to start with `true`, just in case for example the PoW network would be waiting for the checkpoint to arrive (like Advocate would), unable to generate new events and not sending new notifications.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -30,19 +32,38 @@ import scala.annotation.tailrec
...
       highQC: QuorumCertificate[CheckpointingAgreement]
-  ): F[Option[Block]] = ???
+  ): F[Option[Block]] = (
+    for {
+      parent    <- OptionT(getBlock(highQC.blockHash))
+      oldLedger <- OptionT(projectLedger(parent))
+      mempool   <- OptionT.liftF(projectMempool)
+
+      newBody <- OptionT {
+        if (mempool.isEmpty)
```
aakoshh, line 53: It may be worth making this behaviour configurable, in case the interpreter doesn't support notifications for whatever reason.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
```diff
@@ -280,7 +303,7 @@ object CheckpointingService {
       ledgerStorage: LedgerStorage[N],
       blockStorage: BlockStorage[N, CheckpointingAgreement],
       viewStateStorage: ViewStateStorage[N, CheckpointingAgreement],
-      pushCheckpointFn: CheckpointCertificate => F[Unit]
+      interpreterClient: InterpreterRPC[F]
```
aakoshh, line 306: I know this is a draft, but as I [commented](https://github.com/input-output-hk/metronome/pull/53#discussion_r666075541) before I thought `apply` would get the local connection manager and create the client directly, so that it can pass the `ServiceRPC` instance that has access to things like the `mempool` created further down.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: What if another checkpoint notification comes along while creating the block? It's probably not a big issue since it's a really tiny window where it cause a race condition and if it does other federation members should continue including checkpoint candidates, but I'm not entirely happy relying on that. Should we add locking?

aakoshh: Good point! I think the solution is to use [Ref.getAndUpdate](https://github.com/typelevel/cats-effect/blob/c86425cbc88a5d1551ba26a6b76a8cac1d493002/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala#L47) instead of `Ref.get` to access the mempool, and set the "dirty" flag right there when it's read. This would be in `projectMempool` before any further filtering is done, but then if another notification comes in, it's fine because we don't need to update the mempool again during block creation.

rtkaczyk: Related to this is at which point we should save the filtered `Mempool` in its `Ref`, so that we don't lose proposerBlocks that may come along?

What if we had an effectful `projectMempool(hash: Block.Hash): F[Mempool]` and a pure `filterMempool(body: Block.Body)(mempool: Mempool): Mempool`. `projectMempool` uses `filterMempool` but doesn't save the result. Then for each executed block we do `mempoolRef.update(filterMempool(body))`. WDYT?

aakoshh: Also a very good question, something I've been pondering myself over the weeks too. 

I think that removing transactions from the mempool after executing a block is the right thing to do. These transactions either got checkpointed or they ended up in the ledger. We can't do it after creating a block because it may not gather enough votes, and we have to re-propose the same transaction next time. 

There are some potential problems, though, which I mentioned to @fishermanymc. You asked last week how we should deal with proposed transactions that we have already seen somehow, and I said they should be handled by the Interpreter which actually understands the content. For example we could be at height 101, having checkpointed height 100 already (say, every 10th), and someone proposes to include a transaction and height 75. The Interpreter can just reject such block because maybe Advocate only allows uncle blocks that are newer than the last checkpoint, or the one before that (level 100 or 90). 

But what if I'm an honest node and for some reason I just made a block at level 89 and it got into my mempool, how to get rid of it now? Should there be a TTL, in terms of proposal attempts? Or maybe we should have one more message between the Service and the Interpreter, where we actually hand over the mempool to the Interpreter after each committed or executed block to be filtered?

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Yeah, makes sense, but do you mean to include this config now or just make a note of it? Asking, cause I don't believe we actually have any config settings yet. Do you have any concrete vision as to how they should be structured?

aakoshh: What I do in these cases is either:
* Add a constructor argument to indicate that this is an optional behaviour that I can tweak later during application composition.
* When I have two or more of these options, I usually create `case class Config` in the companion object to hold all possible options for this class, perhaps with a `val default = Config(...)` for the default options that I can use as a default parameter value in `apply` and pass to the constructor. For example [ConsensusService.Config](https://github.com/input-output-hk/metronome/blob/pm-3353-quorum-certificate-phase-typed/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala#L580), which gets passed down through `HotStuffService`.

This second option can then be naturally used by the `config` module, so we can have a root config object like this:
```scala
case class CheckpointingConfig(
  network: CheckpointingConfig.Network, 
  checkpointing: CheckpointingService.Config
)
```
Or something along those lines. [RobotConfig](https://github.com/input-output-hk/metronome/pull/42/files#diff-1495f70cdcd12e0a2f23837ea37dd12921eba8d68262c606737f046fb9a0e123) is an example, used throughout [RobotComposition](https://github.com/input-output-hk/metronome/pull/42/files#diff-10b56ac2536b604ca590b5fe6859da38983d0023e5199eb3d5659d1d9cee12ddR72-R75) to pass the values to things that demand it.

rtkaczyk: Oh, thanks, I did not realise we already have something like that. Yeah I would definitely prefer to accumulate those settings using a case class as there will certainly be more. Interpreter timeout being one example. 

I was also thinking that there should be a setting whether to expect any proposer blocks. In case of ETC (`false`) it wouldn't make sense to filter out mempool, especially that it may need to read blocks from storage.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/CheckpointingService.scala
rtkaczyk: Yeah, that sucked, fixed

metronome/hotstuff/consensus/test/src/io/iohk/metronome/hotstuff/consensus/ArbitraryInstances.scala
```diff
@@ -8,13 +8,18 @@ import io.iohk.metronome.hotstuff.consensus.basic.Phase.{
...
 import org.scalacheck.Arbitrary.arbitrary
+import org.scalacheck.rng.Seed
 import org.scalacheck.{Arbitrary, Gen}
 import scodec.bits.ByteVector
 
 trait ArbitraryInstances {
 
   def sample[T: Arbitrary]: T = arbitrary[T].sample.get
 
+  //TODO: rename / remove above?
```
rtkaczyk, line 19: I thought I was going to use it in order to avoid creating extra fixture parameters to pass where `Gen` context is not available, but I did not in the end. Remove?

aakoshh: I'm not exactly sure how to use `sample0`, where to get the `Seed` from, but we can leave it in for rainy days.

aakoshh (approved):
Solid work :+1:

PR #62, opened 2021-07-13
Release/v0.2.0

biandratti:
Bump to v0.2.0


aakoshh:
Hey @biandratti I see you closed this PR. Just to clarify, we came up with this semi-manual release process with Konrad: we use the [version file](https://com-lihaoyi.github.io/mill/mill/Contrib_Modules.html#_version_file) module of `mill` to [remove](https://github.com/input-output-hk/metronome/blob/develop/.circleci/publish.sh#L18) the `-SNAPSHOT` from the version during any build on `master`, so we don't have to do it manually, just do a merge an it's automatically released. 

The only thing we do have to do manually is to update `develop` later to start pushing the next version snapshot instead.

Sorry, I did both of these PRs without waiting for reviews. Thanks for jumping on the opportunity though!

PR #64, opened 2021-07-21
PM-3587: Bump mill version to 0.9.9

aakoshh:
In preparation for https://github.com/NixOS/nixpkgs/pull/130823

I had to separate the tests into `specs` and `props` because apparently since mill 0.9.7 we can only use one test framework per module.


rtkaczyk:
I'm glad that `mill mill.scalalib.GenIdea/idea` still works

rtkaczyk (approved):
Looks good!

PR #65, opened 2021-07-21
FIX: Add ConfigDecoders.booleanDecoder 

aakoshh:
Overrides from `-D` system properties turned Boolean values into String.

Thanks to @biandratti for noticing and the fix.


biandratti (approved):
mm
      /^(  )^\
      \,(..),/
        V~~V 
        
LGTM!

PR #67, opened 2021-07-28
PM-3562: Validate a checkpointing certificate.

aakoshh:
Add `CheckpointCertificate.validate` so that the Interpreter implementations can rely on a function provided by the library to verify the contents.

### DIFF
--- a/metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
+++ b/metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
@@ -56,6 +56,14 @@ object CheckpointCertificateProps extends Properties("CheckpointCertificate") {
         ).map(ByteVector(_)).reduce(_ ++ _)
     )
 
+  /** Extends a chain of blocks `n` more times.
+    *
+    * The `ancestors` are ordered so that the `head` is the most recent block,
+    * the one which should be used as the parent for the extension.
+    *
+    * The return value is reversed so the oldest block is the `head` and the tip is the `last` item.
+    * This can be used to put into the `CheckpointCertificate` in the same order.
+    */
   def buildChain(
       n: Int,
       parentLedger: Ledger,
--- a/metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
+++ b/metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
@@ -16,13 +16,18 @@ import io.iohk.metronome.hotstuff.consensus.basic.{
 }
 import java.security.SecureRandom
 import org.scalacheck._
-import org.scalacheck.Prop.{forAll, propBoolean}
+import org.scalacheck.Prop.{forAll, forAllNoShrink, propBoolean}
 import scodec.bits.ByteVector
+import scala.util.Random
 
 object CheckpointCertificateProps extends Properties("CheckpointCertificate") {
   import ArbitraryInstances._
   import Arbitrary.arbitrary
   import RLPCodecs._
+
+  // Testing with real signatures is rather slow.
+  override def overrideParameters(p: Test.Parameters): Test.Parameters =
+    p.withMinSuccessfulTests(20)
 
   implicit val genFederation: Gen[
     (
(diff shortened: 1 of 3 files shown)


metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
```diff
@@ -117,6 +117,9 @@ object MerkleTree {
       }
     }
 
+    val maxLeafIndex = (1 << proof.siblingPath.length) - 1
```
aakoshh, line 120: This is required so for example on a single element tree a `Proof(4, Nil)` is not accepted, only `Proof(0, Nil)`, otherwise it would mean the `CheckpointCertificate` is malleable by adding arbitrarily large numbers as long as the LSB side that the verification uses is correct.

metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
```diff
@@ -0,0 +1,204 @@
...
+        ).map(ByteVector(_)).reduce(_ ++ _)
+    )
+
+  def buildChain(
+      n: Int,
+      parentLedger: Ledger,
+      ancestors: NonEmptyList[Block.Header]
+  ): Gen[NonEmptyList[Block.Header]] =
+    if (n == 0) {
+      Gen.const(ancestors.reverse)
```
lemastero, line 65: It is unclear for me why we reverse as the last step. Maybe a short description what this Gen achieve?

metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
```diff
@@ -0,0 +1,204 @@
...
+    candidate <- arbitrary[Transaction.CheckpointCandidate]
+    txnsPre   <- arbitrary[Vector[Transaction]]
+    txnsPost  <- arbitrary[Vector[Transaction.ProposerBlock]]
+    txns = txnsPre ++ Vector(candidate) ++ txnsPost
+
+    parent <- arbitrary[Block.Header]
+    ledger <- arbitrary[Ledger].map(_ update txns)
+    block = Block.make(parent, ledger.hash, Block.Body(txns))
+
+    chainLength <- Gen.choose(0, 5)
```
lemastero, line 84: Any particular reason for chain length 5? Would it make sense to use e.g. 20 or something large in general here?

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/CheckpointCertificate.scala
```diff
@@ -31,14 +34,65 @@ case class CheckpointCertificate(
...
+    * and a list of `Block.Header`s leading up to the `QuorumCertifictate` that proves the
+    * BFT agreement over the contents.
+    */
   def construct(
       block: Block,
       headers: NonEmptyList[Block.Header],
       commitQC: QuorumCertificate[CheckpointingAgreement, Phase.Commit]
-  ): Option[CheckpointCertificate] =
+  ): Option[CheckpointCertificate] = {
+    assert(block.hash == headers.head.hash)
```
lemastero, line 47: So we intentionally do not return information that block header do not match (return Option rather Either)?

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
lemastero: That looks like excellent comment to be placed in code.

metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
aakoshh: No reason. Hopefully we won't see chains that big in practice, as it would mean that HotStuff could not reach quorum over 20 views. For the testing purposes, we just have to have a chain with length ranging from 1 to "some", so we can detect any errors mixing up the `head` and `last`, all the in-between headers have to just pass the same rule of parent-child realationship.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/CheckpointCertificate.scala
aakoshh: You're asking why there's an `assert` rather than a "error" return value like `Left`, right?

This assertion is just to make sure we're calling this method with the right data. It would be a programming error if we passed something else, so this just puts into code what's in the comments, to catch simple ordering errors early during development. In practice we should never call it with invalid data, so there's no reason to be prepared for an error to be returned due to these reasons.

metronome/checkpointing/models/props/src/io/iohk/metronome/checkpointing/models/CheckpointCertificateProps.scala
aakoshh: Added docstring.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/CheckpointCertificate.scala
aakoshh: Basically just sanity checking that [these comments](https://github.com/input-output-hk/metronome/blob/f2d70763dc36e37bd18c851e5d1707c255f094f3/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/CheckpointCertificate.scala#L25-L33) are taken into account.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/CheckpointCertificate.scala
```diff
@@ -31,14 +34,65 @@ case class CheckpointCertificate(
...
+      signing: Signing[CheckpointingAgreement]
+  ): Either[String, Validated[Transaction.CheckpointCandidate]] = {
+    def check(
+        message: String
+    )(cond: Boolean): Either[String, Unit] =
+      Either.cond(cond, (), message)
+
+    for {
+      _ <- check(
+        "The headers do not correspond to a chain of parent-child blocks."
```
lemastero, line 73: Would it make sense to say this in positive terms (remove "do not"):
"The headers correspond to a chain of parent-child blocks." etc.

lemastero (commented):
Very nicely structured code. I left a few comments to clarify.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala
aakoshh: Done.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/CheckpointCertificate.scala
aakoshh: I know what you mean, it looks like a statement that already happened, but it becomes an error message where it makes sense this way :-/ 

Scalaz used to have some syntax for validation where it was like `condtion \/ "Whatever failed."`, I'll see if I can reorder it to make it look more natural.

aakoshh: Changed it to `test orError "This failed."` pattern. Does that work better?

aakoshh: I guess alternatively I could use `must be` as a positive term, like `check("The Commit Q.C. must be valid.", signing.validate(commitQC))`.

lemastero: I was not aware that this becomes error message. In that case I am fine with this :)

lemastero (approved):
LGTM 👍

PR #68, opened 2021-08-03
PM-3135: Implement state synchronization in the CheckpointingService

aakoshh:
Adds the boilerplate-y machinery to the `CheckpointingService` to:
* handle `GetStateRequest` and `GetStateResponse` in the `CheckpointinService` 
* implement the `syncState` logic

These are both like-for-like copies from the `RobotService`. Some of the request/response handling has been pulled out from all the classes that had them (`SyncService`, `RobotService`, `InterpreterClient`) so they could be reused in the `CheckpointingService`. I also separated the network message handling into a private class to limit the size of the service and the scope that has to be understood. 

Added tracing as well for the most significant events.

The PR doesn't contain any tests. They'd be rather tedious to write, and I'd rather implement https://jira.iohk.io/browse/PM-3138 in a similar fashion to https://github.com/input-output-hk/metronome/pull/44 to exercise the whole stack under simulated network conditions. We're also a bit short on time, laden with other priorities.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -160,8 +160,8 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
       )
 
       def single(args: String*) = T.command {
-        // ScalaCheck test
-        super.runMain(args.head, args.tail: _*)
+        // ScalaCheck test. Increased verbosity for longer stack traces.
+        super.runMain(args.head, (args.tail ++ Seq("-verbosity", "3")): _*)
       }
     }
   }
--- a/metronome/checkpointing/service/props/src/io/iohk/metronome/checkpointing/service/BlockCreationProps.scala
+++ b/metronome/checkpointing/service/props/src/io/iohk/metronome/checkpointing/service/BlockCreationProps.scala
@@ -81,10 +81,9 @@ object BlockCreationProps extends Properties("BlockCreation") {
         createdBody <- arbitrary[Option[Block.Body]]
 
         hasCheckpointCandidate = createdBody
-          .map(_.transactions.collect {
+          .exists(_.transactions.collect {
             case _: Transaction.CheckpointCandidate =>
           }.nonEmpty)
-          .getOrElse(false)
 
         initialMempool <- arbitrary[Mempool].map(
           _.copy(hasNewCheckpointCandidate = hasCheckpointCandidate).add(
--- a/metronome/checkpointing/service/props/src/io/iohk/metronome/checkpointing/service/BlockCreationProps.scala
+++ b/metronome/checkpointing/service/props/src/io/iohk/metronome/checkpointing/service/BlockCreationProps.scala
@@ -79,8 +79,20 @@ object BlockCreationProps extends Properties("BlockCreation") {
         chain       <- genBlockChain(block, Ledger.empty, min = minChain, max = 5)
         highQC      <- genQC(Phase.Prepare, chain.last.hash)
         createdBody <- arbitrary[Option[Block.Body]]
-        mempool     <- arbitrary[Mempool]
-        initialMempool = mempool.add(chain.flatMap(_.body.proposerBlocks))
+
+        hasCheckpointCandidate = createdBody
+          .map(_.transactions.collect {
+            case _: Transaction.CheckpointCandidate =>
+          }.nonEmpty)
+          .getOrElse(false)
+
+        initialMempool <- arbitrary[Mempool].map(
+          _.copy(hasNewCheckpointCandidate = hasCheckpointCandidate).add(
+            chain.flatMap(_.body.proposerBlocks) ++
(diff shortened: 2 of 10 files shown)


build.sc
```diff
@@ -160,8 +160,8 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
       )
 
       def single(args: String*) = T.command {
-        // ScalaCheck test
-        super.runMain(args.head, args.tail: _*)
+        // ScalaCheck test. Increased verbosity for longer stack traces.
+        super.runMain(args.head, (args.tail ++ Seq("-verbosity", "3")): _*)
```
lemastero, line 164: Nice trick 👍

aakoshh: @rtkaczyk found it :beers:

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/messages/CheckpointingMessage.scala
```diff
@@ -28,4 +28,7 @@ object CheckpointingMessage extends RPCMessageCompanion {
       state: Ledger
   ) extends CheckpointingMessage
       with Response
+
+  implicit val getStatePair =
+    pair[GetStateRequest, GetStateResponse]
```
lemastero, line 33: Interesting encoding 👍  I like the extra safety from enforcing proper response type.

metronome/core/src/io/iohk/metronome/core/messages/RPCSupport.scala
```diff
@@ -0,0 +1,110 @@
...
+    protected def receiveResponse[Res <: Response](
+        response: Res
+    ): F[Unit] =
+      rpcTracker.complete(response).flatMap {
+        case Right(ok) =>
+          responseIgnored(response, None).whenA(!ok)
+        case Left(ex) =>
+          responseIgnored(response, Some(ex))
+      }
+  }
```
lemastero, line 109: Remote and Local seems to be very similar to each other. Yet I do not see ATM how one could by more DRY. Possibly Local is just a special case of Remote when `to` argument is self.

metronome/checkpointing/service/props/src/io/iohk/metronome/checkpointing/service/BlockCreationProps.scala
```diff
@@ -79,8 +79,20 @@ object BlockCreationProps extends Properties("BlockCreation") {
...
         highQC      <- genQC(Phase.Prepare, chain.last.hash)
         createdBody <- arbitrary[Option[Block.Body]]
-        mempool     <- arbitrary[Mempool]
-        initialMempool = mempool.add(chain.flatMap(_.body.proposerBlocks))
+
+        hasCheckpointCandidate = createdBody
+          .map(_.transactions.collect {
+            case _: Transaction.CheckpointCandidate =>
+          }.nonEmpty)
+          .getOrElse(false)
```
lemastero, line 87:
```scala
hasCheckpointCandidate = createdBody.exists(_.transactions.collect {
  case _: Transaction.CheckpointCandidate =>
}.nonEmpty)
```

lemastero (approved):
Nice one 👍

metronome/core/src/io/iohk/metronome/core/messages/RPCSupport.scala
aakoshh: My first attempt was where `Local` just inherited `Remote` using `Unit` for `PKey` and passing `()`, but the problem was that with inheritance the classes that used them got both versions of the `sendRequest` methods, which didn't feel right. 

I thought about not using inheritance but some component instead; after all I can't expect that none of the classes that would use this have no other base class. Then I could use a separate `Local` and `Remote` trait, and some common base class to hide the details. I think I didn't want to have to declare all these type parameters twice.

I'll have another look at it.

metronome/checkpointing/service/src/io/iohk/metronome/checkpointing/service/messages/CheckpointingMessage.scala
aakoshh: Yeah and it lets the compiler figure out the return value. Even better would be to use session types, but I went too much into the direction of queue processing.

metronome/core/src/io/iohk/metronome/core/messages/RPCSupport.scala
aakoshh: On second thought, I removed the `Local` and just pass `()` as the address in `InterpreterClient`. Less annoyance than having the duplicated code.

aakoshh: And I suppose you can always create a field with an implementation of the class if inheritance is not an option, an inner class, so that's not a problem. Although the fact that the methods are `protected` could be inconvenient.

PR #69, opened 2021-08-19
PM-3699: Option to return transactions to be purged during block creation

aakoshh:
Adds the ability for the Interpreter to signal to the Service that some transactions in the mempool can be unconditionally forgotten.

Normally we'd keep the transactions in the mempool until a block that contains them is executed, however, some transactions may be there that can never be included in a block. Since `CreateBlockBodyRequest` contains the `mempool`, `CreateBlockBodyResponse` can be used to signal that some of the content has been inspected and is no longer interesting.

### DIFF
--- a/metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
+++ b/metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
@@ -26,7 +26,7 @@ trait InterpreterRPC[F[_]] {
   def createBlockBody(
       ledger: Ledger,
       mempool: Seq[Transaction.ProposerBlock]
-  ): F[Option[(Block.Body, Set[Transaction.ProposerBlock])]]
+  ): F[Option[InterpreterRPC.CreateResult]]
 
   def validateBlockBody(
       blockBody: Block.Body,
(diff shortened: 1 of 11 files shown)


metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
```diff
@@ -26,7 +26,7 @@ trait InterpreterRPC[F[_]] {
   def createBlockBody(
       ledger: Ledger,
       mempool: Seq[Transaction.ProposerBlock]
-  ): F[Option[Block.Body]]
+  ): F[Option[(Block.Body, Set[Transaction.ProposerBlock])]]
```
rtkaczyk, line 29: I wouldn't mind a doc string explaining the return type, as on its own it's not very clear

rtkaczyk (approved):
:+1:

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/InterpreterRPC.scala
aakoshh: Yeah, I was wondering if I should introduce a custom case class or at least a type alias for it, because it got tiring to type it in.  Just when I typed in `BlockBodyWithPurgeFromMempool` I thought that's a mouthful. I'll have another go at it.

aakoshh: Added the type alias and some helper functions, thanks for calling this out.

span14:
I see. We give that suggestion based on the decision that the service has the mempool and the interpreter doesn't. But your solution also works, we can adapt to it by having the interpreter asking for the checkpointing candidate in the main chain.

PR #70, opened 2021-08-23
Added failover codecs for json array override support

dmitry-worker:
System property override like
```
./bin/node -Dblockchain.testnet.bootstrap-nodes.0='enode://asdfasdf'
```
Did not actually override an element, but turned the array into `JsonObject` instead!
```
"bootstrap-nodes":{"0":"enode://asdfasdf"}
```

In order to handle those Seq, List, Vector codec failovers were introduced.
Gaps between keys are **no longer** ignored, so keys like `{1,0,2}` will be accepted and values sorted, but `{3,5,2}` will fail the decoder with a corresponding message

Testcases were provided as well

### DIFF
--- a/metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
+++ b/metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
@@ -53,30 +53,77 @@ class ConfigParserSpec
     json.noSpaces shouldBe """{"nestedStructure":{"barBaz":{"spam":"eggs"},"foo":10}}"""
   }
 
-  "seq,list,vector decoders" should "handle JSON Object with indices successfully" in {
+  "seq,list,vector decoders" should "handle JSON Object with sequential indices successfully" in {
     import ConfigParserSpec.TestConfigWithJsonArray._
     val withArray = """{"field":["valueA", "valueB", "valueC"]}"""
     val withObject = """{"field":{"2":"valueC", "0": "valueA", "1":"valueB"}}"""
+    val expected = Vector("valueA", "valueB", "valueC")
     val aConf = ConfigFactory.parseString(withArray)
     val oConf = ConfigFactory.parseString(withObject)
 
-    //seqs
+    // seqs
     val aSeq = ConfigParser.parse[TestConfigWithSeq](aConf.root)
     val oSeq = ConfigParser.parse[TestConfigWithSeq](oConf.root)
-    aSeq shouldBe a[Right[_,_]]
+    inside(aSeq) { case Right(value) => value.field.toVector shouldBe expected }
     aSeq shouldEqual oSeq
 
-    //lists
+    // lists
     val aList = ConfigParser.parse[TestConfigWithList](aConf.root)
     val oList = ConfigParser.parse[TestConfigWithList](oConf.root)
-    aList shouldBe a[Right[_,_]]
+    inside(aList) { case Right(value) => value.field.toVector shouldBe expected }
     aList shouldEqual oList
 
-    //vectors
+    // vectors
     val aVector = ConfigParser.parse[TestConfigWithVector](aConf.root)
     val oVector = ConfigParser.parse[TestConfigWithVector](oConf.root)
-    aVector shouldBe a[Right[_,_]]
+    inside(aVector) { case Right(value) => value.field shouldBe expected }
     aVector shouldEqual oVector
+  }
+
+  "seq,list,vector decoders" should "fail on JSON Object key gaps but allow empty object" in {
+    import ConfigParserSpec.TestConfigWithJsonArray._
+
+    val withGaps1 = """{"field":{"2":"valueC", "0":"valueA"}}"""
+    val withGaps2 = """{"field":{"2":"valueA"}}"""
+    val withDupes = """{"field":{"2":"valueA", "2":"valueB"}}"""
+    val withEmpty = """{"field":{}}"""
+    val gapsConf1 = ConfigFactory.parseString(withGaps1)
+    val gapsConf2 = ConfigFactory.parseString(withGaps2)
+    val dupesConf = ConfigFactory.parseString(withDupes)
+    val emptyConf = ConfigFactory.parseString(withEmpty)
+
+    // seqs
+    val gapsSeq1 = ConfigParser.parse[TestConfigWithSeq](gapsConf1.root)
+    val gapsSeq2 = ConfigParser.parse[TestConfigWithSeq](gapsConf2.root)
+    val dupesSeq = ConfigParser.parse[TestConfigWithSeq](dupesConf.root)
+    val emptySeq = ConfigParser.parse[TestConfigWithSeq](emptyConf.root)
+
+    inside(gapsSeq1) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 2) sequence, but got {0, 2}") }
+    inside(gapsSeq2) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(dupesSeq) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(emptySeq) { case Right(success) => success.field shouldBe empty }
+
+    // lists
+    val gapsList1 = ConfigParser.parse[TestConfigWithList](gapsConf1.root)
+    val gapsList2 = ConfigParser.parse[TestConfigWithList](gapsConf2.root)
+    val dupesList = ConfigParser.parse[TestConfigWithList](dupesConf.root)
+    val emptyList = ConfigParser.parse[TestConfigWithList](emptyConf.root)
+
+    inside(gapsList1) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 2) sequence, but got {0, 2}") }
+    inside(gapsList2) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(dupesList) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(emptyList) { case Right(success) => success.field shouldBe empty }
+
+    // vectors
+    val gapsVector1 = ConfigParser.parse[TestConfigWithVector](gapsConf1.root)
+    val gapsVector2 = ConfigParser.parse[TestConfigWithVector](gapsConf2.root)
+    val dupesVector = ConfigParser.parse[TestConfigWithVector](dupesConf.root)
+    val emptyVector = ConfigParser.parse[TestConfigWithVector](emptyConf.root)
+
+    inside(gapsVector1) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 2) sequence, but got {0, 2}") }
+    inside(gapsVector2) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(dupesVector) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(emptyVector) { case Right(success) => success.field shouldBe empty }
   }
 
   "withEnvVarOverrides" should "overwrite keys from the environment" in {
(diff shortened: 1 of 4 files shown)


metronome/config/src/io/iohk/metronome/config/ConfigDecoders.scala
```diff
@@ -35,6 +35,40 @@ object ConfigDecoders {
...
+    vectorDecoder.map(_.toSeq) // returns vector itself, actually
+  }
+
+  implicit def listDecoder[A](implicit aDecoder: Decoder[A]): Decoder[List[A]] = {
+    vectorDecoder.map(_.toList)
+  }
+
+  implicit def vectorDecoder[A](implicit aDecoder: Decoder[A]): Decoder[Vector[A]] = {
+    Decoder.decodeVector[A] or Decoder[Map[Int, A]].map { srcMap =>
+      srcMap.toVector.sortBy(_._1).map(_._2)
```
aakoshh, line 68: So we discussed on Slack that this could check that indices actually go from `0..n` in a contiguous fashion, so for example the user does not live under the impression that they have overridden just item `arr.2` when they have in fact reduced their array to `[arr.2]` and discarded all other default items. 

You seemed to agree there.

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -53,6 +53,32 @@ class ConfigParserSpec
...
+    import ConfigParserSpec.TestConfigWithJsonArray._
+    val withArray = """{"field":["valueA", "valueB", "valueC"]}"""
+    val withObject = """{"field":{"2":"valueC", "0": "valueA", "1":"valueB"}}"""
+    val aConf = ConfigFactory.parseString(withArray)
+    val oConf = ConfigFactory.parseString(withObject)
+
+    //seqs
+    val aSeq = ConfigParser.parse[TestConfigWithSeq](aConf.root)
+    val oSeq = ConfigParser.parse[TestConfigWithSeq](oConf.root)
+    aSeq shouldBe a[Right[_,_]]
```
aakoshh, line 66: It would be good to actually check that the parsed result contains what we expect. Could be just an empty list, and the test wouldn't highlight it.

aakoshh (commented):
Thanks you for the contribution! Could you add your name to the list of developers in `build.sc` :partying_face: ?

I'd like to see a test under the [system property overrides](https://github.com/input-output-hk/metronome/blob/4c51905b016447a7f008a75f5c0109a9058593de/metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala#L150) so we actually demonstrate the whole behaviour, rather than rely on what you discovered with logging.

It should also demonstrate that an existing array of N items is overridden completely by M items specified in the system properties. If you could tweak the test to take a list of overrides rather than use `withProperty` in a nested fashion that would be great!

And I think that going with the conservative version and rejecting the array indices unless they are contiguous would avoid some potential misunderstandings.

biandratti (commented):
Can you add a new case in the withEnvVarOverrides test, adding a new environment variable to override an element in the array?

In this way, both convictions should work, right?
This means that TEST_NETWORK_BOOTSTRAP_0 and TEST_NETWORK_BOOTSTRAP.0 could be used to override an element in an array.

aakoshh:
> This means that TEST_NETWORK_BOOTSTRAP_0 and TEST_NETWORK_BOOTSTRAP.0 could be used to override an element in an array. That's right?

I didn't think `TEST_NETWORK_BOOTSTRAP.0` would work right now because be the [extend applied on indexes](https://github.com/input-output-hk/metronome/blob/4c51905b016447a7f008a75f5c0109a9058593de/metronome/config/src/io/iohk/metronome/config/ConfigParser.scala#L167) uses `_` to construct the env var name. Even with the built in method of overriding with env vars, HOCON expects it like `export CONFIG_FORCE_items_0=a`, so I don't think we should have `.` in env var names. At least I can't remember seeing it used.

Note thought that env var ovverrides unfortunately work different from the system properties: with system properties you can override the whole array, whereas with env vars you can only override existing items, because all it does is traverse the JSON document that represents the config map.

biandratti (approved):
mm
      /^(  )^\
      \,(..),/
        V~~V 
        
LGTM!

aakoshh:
The HOCON `CONFIG_FORCE_` env vars are a [bit more nuanced](https://github.com/lightbend/config#optional-system-or-env-variable-overrides) than what I implemented, they support `_`, `__` and `___` in the path, meaning `.`, `-` and `_`, respectively. With that they could restore the exact object structure that you can do with system properties, so you can construct entire sub-structures in the JSON tree. Whereas with the simpler convention that I did I can't figure out whether `A_B_C=1` should be `{"a": {"b-c": 1}}` or `{"a": {"b": {"c": 1}}}` for example, therefore I can't create new items in an empty JSON array, I can just override what exists. 

But of course anyone is free to use the `CONFIG_FORCE_` version, it just requires more familiarity with how it works and how the config file looks like.

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -53,6 +53,79 @@ class ConfigParserSpec
...
+    // vectors
+    val gapsVector1 = ConfigParser.parse[TestConfigWithVector](gapsConf1.root)
+    val gapsVector2 = ConfigParser.parse[TestConfigWithVector](gapsConf2.root)
+    val dupesVector = ConfigParser.parse[TestConfigWithVector](dupesConf.root)
+    val emptyVector = ConfigParser.parse[TestConfigWithVector](emptyConf.root)
+
+    inside(gapsVector1) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 2) sequence, but got {0, 2}") }
+    inside(gapsVector2) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(dupesVector) { case Left(Right(err)) => err.message shouldBe (s"Expected [0, 1) sequence, but got {2}") }
+    inside(emptyVector) { case Right(success) => success.field shouldBe empty }
```
aakoshh, line 126: Would it be possible to reduce the duplication here? Maybe use `behave like`? Or just test this with one of the types, relying on the other test to demonstrate that they are equivalent?

aakoshh: Or perhaps `TableDrivenTests`?

metronome/config/src/io/iohk/metronome/config/ConfigDecoders.scala
```diff
@@ -35,6 +35,47 @@ object ConfigDecoders {
...
+
+  implicit def vectorDecoder[A](implicit aDecoder: Decoder[A]): Decoder[Vector[A]] = {
+    Decoder.decodeVector[A] or Decoder[Map[Int, A]].emap { srcMap =>
+      val sorted = srcMap.toVector.sortBy(_._1)
+      if (sorted.isEmpty || sorted.size == sorted.last._1 + 1) {
+        // no gaps, because key uniqueness is guaranteed by Map
+        Right(sorted.map(_._2))
+      } else {
+        val keyPrintout = sorted.map(_._1).mkString("{", ", ", "}")
+        Left(s"Expected [0, ${sorted.size}) sequence, but got $keyPrintout")
```
aakoshh, line 74: Nice :heart:

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
aakoshh: Also some helper function could be used to avoid having to have 4 different variables for parsed configs, something like 
```scala
def check[T : Decoder](s: String)(t: T => Assertion): Unit = ???
```

metronome/config/src/io/iohk/metronome/config/ConfigDecoders.scala
dmitry-worker: Yep, now I cleaned it up, made some tests that detect such override mistakes. Thanks!

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
dmitry-worker: That's done!

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
dmitry-worker: Added checkDecode helper function and split this test into 3 separate testcases!

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -53,6 +55,72 @@ class ConfigParserSpec
...
+    val gapsConf2 = ConfigFactory.parseString("""{"field":{"2":"valueA"}}""")
+
+    checkDecoding[TestConfigWithSeq](gapsConf1, Left(s"Expected [0, 2) sequence, but got {0, 2}"))
+    checkDecoding[TestConfigWithSeq](gapsConf2, Left(s"Expected [0, 1) sequence, but got {2}"))
+
+    checkDecoding[TestConfigWithList](gapsConf1, Left(s"Expected [0, 2) sequence, but got {0, 2}"))
+    checkDecoding[TestConfigWithList](gapsConf2, Left(s"Expected [0, 1) sequence, but got {2}"))
+
+    checkDecoding[TestConfigWithVector](gapsConf1, Left(s"Expected [0, 2) sequence, but got {0, 2}"))
+    checkDecoding[TestConfigWithVector](gapsConf2, Left(s"Expected [0, 1) sequence, but got {2}"))
```
aakoshh, line 100:
```suggestion
    def check[T : Decoding] = {
      checkDecoding[T](gapsConf1, Left(s"Expected [0, 2) sequence, but got {0, 2}"))
      checkDecoding[T](gapsConf2, Left(s"Expected [0, 1) sequence, but got {2}"))
    }
    
    check[TestConfigWithSeq]
    check[TestConfigWithList]
    check[TestConfigWithVector]
```

Sorry, don't want to be obtuse but why repeat the same args.

dmitry-worker: This is much better anyway :)

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -53,6 +55,72 @@ class ConfigParserSpec
...
+    import ConfigParserSpec.TestConfigWithJsonArray._
+    // format: off
+    val emptyConf = ConfigFactory.parseString("""{"field":{}}""")
+    checkDecoding[TestConfigWithSeq](emptyConf, Right(_.field shouldBe empty))
+    checkDecoding[TestConfigWithList](emptyConf, Right(_.field shouldBe empty))
+    checkDecoding[TestConfigWithVector](emptyConf, Right(_.field shouldBe empty))
+    // format: on
+  }
+
+  "seq,list,vector decoders" should "succeed on JSON Object with duplicated keys keeping the last value" in {
```
aakoshh, line 114: Thank you for making an explicit test case for duplicated keys. 

I wanted to say that it would be more helpful to reject such input, but I suppose by the time we get a JSON object `circe` has already collapsed it into `Map` and we can't tell that there were any duplicate values, right?

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -124,35 +192,126 @@ class ConfigParserSpec
...
+      k -> Option(System.setProperty(k, v))
+    }.toMap
     try {
       ConfigFactory.invalidateCaches()
       thunk
     } finally {
-      maybeCurrent.foreach(System.setProperty(key, _))
+      current.foreach {
+        case (k, Some(v)) => System.setProperty(k, v)
+        case (k, _)       => System.clearProperty(k)
```
aakoshh, line 276: Nice one :+1:

aakoshh (approved):
Thanks @dmitry-worker , great stuff :beers:

aakoshh:
A bit late now, but a way to deal with repeating every thrice would have been to use [behave like](https://www.scalatest.org/user_guide/sharing_tests) such as
```scala
"List" should behave like vectorWithFallback[TestClassWithList](_.fields)
"Vector" should behave like vectorWithFallback[TestClassWithVector](_.fields)

def vectorWithFallback[T : Decoder](T => Vector[String]) {
  it should "work for empty array" in ...
  it should "return error for gaps" in ...
}
```
Just more test cases printed, but you see which failed.

PR #74, opened 2021-08-29
PM-3140: Advocate messages spec

span14:
This PR creates message type for communication between interpreter and chain based on [inter-module-messages](https://input-output.atlassian.net/wiki/spaces/MID/pages/2559901764/Metronome+Inter-Module+Messaging+Specifications) diagram.


metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/messages/InterpreterMessage.scala
```diff
@@ -167,9 +180,135 @@ object InterpreterMessage extends RPCMessageCompanion {
...
+
+  /** The interpreter requests the chain for block at target height in order to construct
+    * the service block
+    */
+  case class CheckpointBlockRequest(
+      requestId: RequestId,
+      blockNumber: BigInt
+  ) extends InterpreterMessage
+      with Request
+      with FromInterpreter
```
aakoshh, line 193: If you read the definition of the `InterpreterMessage`, it says:
```scala
/** Messages exchanged between the Checkpointing Service
  * and the local Checkpointing Interpreter.
  */
sealed trait InterpreterMessage { self: RPCMessage => }
```

You cannot extend it to also contain messages that go between the ETC chain and the interpreter, the two are completely separate things.

metronome/checkpointing/interpreter/src/io/iohk/metronome/checkpointing/interpreter/messages/InterpreterMessage.scala
```diff
@@ -167,9 +180,135 @@ object InterpreterMessage extends RPCMessageCompanion {
...
+
+  /** The chain replies to @CheckpointBlockRequest with block metadata */
+  case class CheckpointBlockResponse(
+      requestId: RequestId,
```
aakoshh, line 197: I think there is no need to model the interaction between the ETC chain and the interpreter with `RPCMessage` machinery, i.e. the `requestId`. That's because this is for _Remote_ Procedure Calls and the interpreter, according to the current design, will run inside the ETC process, in other words it's a _local_ call, so it can simply be a `Task` for example, modeled as a trait, similar to the `InterpreterRPC` and `ServiceRPC` interfaces.

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/ETCBlock.scala
```diff
@@ -0,0 +1,36 @@
...
+
+/** ETCBlock is a simpler version of original ETC block in the chain that
+  * supports the running of the checkpointing interpreter.
+  */
+case class ETCBlock(
+    header: ETCBlock.Header,
+    body: ETCBlock.Body
+) {
+  def hash: Hash = header.hash
+}
```
aakoshh, line 13: ETC already has its own block definition. I don't think we should see chain specific stuff in `checkpointing.models`. Whaver is required to implement the interpreter should live in https://github.com/input-output-hk/mantis

Only the most generic code should be added to the `checkpointing` module that works with everything. I don't want it to end up as a union of all possible chains we can deal with.

aakoshh (changes requested):
After submitting your PR, please check whether you see the expected number of commits and the changes are all yours. As you see it's a mix of your commits, Dmitry's and mine. First you should rebase your changes on the current `develop`. 

I also think that this PR is not needed, in my opinion there are a few misunderstandings here. 

1. The message passing model for ETC is just that, a model. The interpreter will run _inside_ ETC, so it has full access to the database, it doesn't need RPC messages; those are designed for things that need to go over the wire, hence the `requestId` to match up responses with requests. Locally you have much more convenient tools at your disposal.
2.  You don't have to create code for the ETC interpreter, that will be done by @dzajkowski and his team.

aakoshh:
To implement the Advocate interpreter, add a reference to this library in the project where the interpreter is going to be implemented. 

```
"io.iohk" %% "metronome-checkpointing-interpreter" % "0.2.0"
```

metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/ETCBlock.scala
```diff
@@ -0,0 +1,36 @@
...
+  case class Header(
+      parentHash: Hash,
+      hash: Hash,
+      blockNumber: BigInt
+  )
+
+  object Header
+
+  case class Body(
+      transactions: Seq[Hash]
```
aakoshh, line 29: Why would you need the transactions?

aakoshh:
If, in your implementation of the Advocate interpreter, you really want to use messaging internally between the interpreter and the PoW chain, I would recommend having a look at the [Pipe](https://github.com/input-output-hk/metronome/blob/develop/metronome/core/src/io/iohk/metronome/core/Pipe.scala) abstraction. I use it for communication between the [ConsensusService](https://github.com/input-output-hk/metronome/blob/develop/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/ConsensusService.scala#L55) and the [SyncService](https://github.com/input-output-hk/metronome/blob/develop/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/SyncService.scala#L49). 

If you look at its specialisation, the [SyncPipe](https://github.com/input-output-hk/metronome/blob/develop/metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/pipes/SyncPipe.scala) you see that there are requests and responses, but the response has the request itself, not an ID, so you don't need an instance of an `RPCTracker` to match them up, you just use pattern matching. This is something people do with Akka actors as well: if you need some context to process the response, just send it on a round trip. 

The `Pipe` is used in situation where I don't necessarily expect an answer to all the messages. If I did, I'd use a simple `trait` with methods that return `Task` for example. The `Pipe` also facilitates handling messages one by one.

Since it's part of the `metronome-core` library, if you add a reference to `metronome-interpreter`, it will bring them in as a dependency, but you can also add an explicit reference. Like I said this is an option if you want to implement the interpreter with explicit message passing. It's similar to using Akka actors, but it's typed and is just a two way channel, the consuming and sending is all manual, and is only meant to be used between two parties.

PR #75, opened 2021-09-01
[PM-3144, PM-3106]: Checkpointing app main

aakoshh:
Configuration and entry point for the checkpointing application.

- [x] Configuration
- [x] CLI entry for generating a pair of keys
- [x] Composition root
- [x] Read private key from a file

No tests, that was supposed to be in the next PR with the simulated chain events. It's mostly a carbon copy of the robot example.

Added examples to the README about calling it. There are two entry points:
* `keygen` generates a pair of ECDSA keys and prints them to the console in JSON
* `service` runs the Checkpointing Service which will try to connect to the Interpreter.


metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockPruning.scala
```diff
@@ -0,0 +1,36 @@
...
+      viewStateStorage: ViewStateStorage[N, A],
+      blockHistorySize: Int
+  ): KVStore[N, Unit] = {
+    for {
+      // Always keep the last executed block.
+      lastExecutedBlock <- viewStateStorage.getLastExecutedBlockHash.lift
+      pathFromRoot      <- blockStorage.getPathFromRoot(lastExecutedBlock).lift
+
+      // Keep the last N blocks.
+      pruneable = pathFromRoot.reverse
```
rtkaczyk, line 21: I believe the name is somewhat confusing, they're actually `nonPruneable`, aren't they?

metronome/checkpointing/models/specs/src/io/iohk/metronome/checkpointing/models/CheckpointingSigningSpec.scala
```diff
@@ -0,0 +1,85 @@
...
+  LeaderSelection,
+  ViewNumber
+}
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import java.security.SecureRandom
+import scodec.bits.ByteVector
+
+/** A single positive case spec to test type interoperability.
```
rtkaczyk, line 17: We may want to update this comment

metronome/checkpointing/app/src/io/iohk/metronome/checkpointing/app/CheckpointingComposition.scala
```diff
@@ -0,0 +1,572 @@
...
+  protected def makeConsensusTracers: CTS =
+    CheckpointingConsensusTracers.consensusHybridLogTracers
+
+  protected def makeSyncTracers: STS =
+    CheckpointingSyncTracers.syncHybridLogTracers
+
+  protected def makeServiceTracer =
+    CheckpointingServiceTracers.serviceEventHybridLogTracer
+
+  protected def makeConnectionProvider[M: Codec](
```
rtkaczyk, line 180: Could we add explicit type annotations for the protected methods?

metronome/examples/src/io/iohk/metronome/examples/robot/app/RobotComposition.scala
```diff
@@ -132,6 +133,8 @@ trait RobotComposition {
       viewStateStorage <- makeViewStateStorage(genesis)
       stateStorage     <- makeStateStorage(config, genesisState)
 
+      _ <- makeBlockPruner(config, blockStorage, viewStateStorage)
```
rtkaczyk, line 136: Are you just tidying up or is there a functional reason to move it?

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockPruning.scala
aakoshh: No these are definitely pruneable. Maybe the comment was confusing :thinking: 

I'm not sure why I haven't just used `pathFromRoot.dropRight(blockHistorySize)` here, but the point is that we can prune all but the last N items. I'm going to do that by calling `pruneNonDescendants` on the last item of the `pruneable` list.

aakoshh: Tried to add more descriptive comments.

metronome/checkpointing/app/src/io/iohk/metronome/checkpointing/app/CheckpointingComposition.scala
aakoshh: Sure, I thought it would be self-describing and didn't want to add noise, but if you think it helps I'll add those.

metronome/examples/src/io/iohk/metronome/examples/robot/app/RobotComposition.scala
aakoshh: The two composition classes were so similar in structure that I thought it will help anyone looking at both to see that the same thing appears in both at the same place, which is right after the storage components have been created. This should be less surprising than finding it further down.

metronome/hotstuff/service/src/io/iohk/metronome/hotstuff/service/storage/BlockPruning.scala
```diff
@@ -0,0 +1,37 @@
...
+      blockHistorySize: Int
+  ): KVStore[N, Unit] = {
+    for {
+      // Always keep the last executed block.
+      lastExecutedBlock <- viewStateStorage.getLastExecutedBlockHash.lift
+      pathFromRoot      <- blockStorage.getPathFromRoot(lastExecutedBlock).lift
+
+      // Everything but the last N blocks in the chain leading up to the
+      // last executed block can be pruned.
+      pruneable = pathFromRoot.dropRight(blockHistorySize)
```
rtkaczyk, line 23: Ok, I guess the double reversal confused me. But also the fact that with this we are actually keeping `N + 1` blocks, aren't we?

What about something like:
```maybeNewRoot = pathFromRoot.reverse.lift(blockHistorySize - 1)```

aakoshh: Yeah, it's N+1, I thought that was close enough :)

Yours is a nice solution too. I thought having this interim `pruneable` variable is a good mental anchor. But we can go with your version.

aakoshh: Done.

metronome/checkpointing/app/src/io/iohk/metronome/checkpointing/app/CheckpointingComposition.scala
aakoshh: Done.

metronome/checkpointing/models/specs/src/io/iohk/metronome/checkpointing/models/CheckpointingSigningSpec.scala
aakoshh: Done.

rtkaczyk (approved):
I like how it all falls in to place ❤️

PR #76, opened 2021-09-06
Enable nix flake support

jonringer:
Just building the example right now, but can be modified later to build other applications.

cc @aakoshh

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,43 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1628572867,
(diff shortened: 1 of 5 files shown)


aakoshh:
@jonringer thank you so much for putting this together! I have very little clue of what I'm looking at, though. Do you think some README entries would help here? What do we have to do to be able to use the Bitte tool, which required the Nix build, to deploy some nodes for benchmarking?

lemastero (approved):
LGTM

jonringer:
> @jonringer thank you so much for putting this together! I have very little clue of what I'm looking at, though. Do you think some README entries would help here? What do we have to do to be able to use the Bitte tool, which required the Nix build, to deploy some nodes for benchmarking?

I think instructions on how to deploy it using just mill would be more beneficial. Easy for me to "nixify" it.

PR #77, opened 2021-09-09
Escape special characters in building a JSON

biandratti:
The goal is to escape special characters.
In this direction, we call the "quoteString" method from typesafe config which is responsible for getting this result.


metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -22,6 +22,24 @@ class ConfigParserSpec
...
+    val json = ConfigParser.withCamelCase(orig)
+
+    val env = Map(
+      "JAVA_HOME" -> """C:\Program Files\java\jdk-11""""
+    )
+
+    val result = ConfigParser.withEnvVarOverrides(json, "", env)
+
+    inside(result) { case Right(json) =>
+      json.noSpaces shouldBe """{"javaHome":"C:\\Program Files\\java\\jdk-11\""}"""
```
dmitry-worker, line 39: It seems a bit strange when you override 
`C:\\Program Files\\java\\jdk-8\`   <- see the single backslash at the end 
with
`C:\Program Files\java\jdk-11` <- no backslash at the end
but expect 
`C:\\Program Files\\java\\jdk-11\`  <- with single backslash at the end

Or maybe I couldn't get what does the test check.
If it is testing only json escape symbols, then perhaps we can test different strings, without a config for these sequences: 
```
      char = unescaped /
          escape (
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
              %x75 4HEXDIG )  ; uXXXX                U+XXXX

      escape = %x5C              ; \

      quotation-mark = %x22      ; "

      unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
```

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -22,6 +22,24 @@ class ConfigParserSpec
     json.noSpaces shouldBe """{"nested-structure":{"bar_baz":{"spam":"eggs"},"foo":10}}"""
   }
 
+  "toJsonEscape" should "parse and escape escape.conf to JSON" in {
```
dmitry-worker, line 25:
```suggestion
  it should "deal with escape characters according to rfc7159 (section-7)" in {
```

dmitry-worker (commented):
Looks good but maybe the testcase should be more informative about what it does.

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
aakoshh:
```
val conf = ConfigFactory.parseString(
      """{"java_home":"C:\\Program Files\\java\\jdk-8\""}"""
                                                      ^
```

It's not intuitive but there _is_ a `"` in the original, which is escaped as `\"`, followed by the actual `"` that closes the string, so the `java_home` path has a double quote in it.

The override _also_ has a double quote in it, but this time it doesn't need escaping because of triple quoting:
```
    val env = Map(
      "JAVA_HOME" -> """C:\Program Files\java\jdk-11""""
                                                    ^
```

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
aakoshh: I would actually move this further down to appear under the `withEnvVarOverrides` section.

aakoshh (commented):
Thanks a lot for this fix! 

Looks good, although I'd consider removing the `"` from the path in the test since it has proven to be confusing. Maybe add another example where a double quote is more natural, like a field that is an encoded JSON string? Can't think of why we'd need one of those but I don't have a better idea.

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
biandratti: Thanks for the recommendation, I applied clearer examples

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
biandratti: Sure, now it is done

dmitry-worker (approved):
🦇 Great job! LGTM 🦇

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
```diff
@@ -175,6 +175,41 @@ class ConfigParserSpec
...
+
+    val env = Map(
+      "QUOTATION_MARK"  -> """"""",
+      "REVERSE_SOLIDUS" -> """\""",
+      "SOLIDUS"         -> """/""",
+      "BACKSPACE"       -> """\b""",
+      "FORM_FEED"       -> """\f""",
+      "LINE_FEED"       -> """\n""",
+      "CARRIAGE_RETURN" -> """\r""",
+      "TAB"             -> """\t""",
```
aakoshh, line 201: These examples aren't quite right. You see if you want to actually place a newline in a string, you should not use triple quotes.

```shell
scala> println(""" foo \n bar """)
 foo \n bar 

scala> println(" foo \n bar ")
 foo 
 bar 
```

The point of triple quotes is that we don't have to escape special characters inside them, they are as-is. 

When you see this JSON: `"lineFeed":"\\n"` that's not a line feed but the characters `\` and `n` in a string. A JSON string with a line feed would look like `"lineFeed":"\n"`. 

That was the error we got: `"javaHome": "C:\Program Files\Java"` had `\P` which looked like an escape of a special character, but `\P` doesn't exist. So to have special chars, they should not be escaped, hence no need to triple quote them, and no need to escape in the assertion either.

aakoshh (approved):
Top man :clap:

metronome/config/specs/src/io/iohk/metronome/config/ConfigParserSpec.scala
biandratti: That's true, I overlooked it.
Regarding the assertion, I keep the message such string with a triple quote. Because the idea is to scape the special characters, and that is working and testing with this use case.
Thanks for note this important point

aakoshh:
Actually, since this is a bugfix, it would be better to rebase this on the current `master` branch, bump the patch version in the PR, merge into `master`, then open a PR from `master` to `develop`.

PR #82, opened 2021-09-27
Publishing for RocksDB and Storage

dmitry-worker:
This very simple PR is meant to turn `rocksdb` and `storage` modules into published libraries

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -434,7 +434,7 @@ class MetronomeModule(val crossScalaVersion: String) extends CrossScalaModule {
     override val description =
       "A functional wrapper for RocksDB"
 
-    override def moduleDeps: Seq[JavaModule] =
+    override def moduleDeps: Seq[PublishModule] =
       Seq(storage)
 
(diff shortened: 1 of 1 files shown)


aakoshh (approved):
Cool! As I said, if you feel like trying to upgrade RocksDB to give transaction isolation another go, you're more then welcome! The [testbed](https://github.com/input-output-hk/metronome/blob/develop/metronome/rocksdb/props/src/io/iohk/metronome/rocksdb/RocksDBStoreProps.scala) in the `rocksdb` module should be helpful, and then it's just a matter of calling a different method from the [KVStoreRunner implementation](https://github.com/input-output-hk/metronome/blob/develop/metronome/checkpointing/app/src/io/iohk/metronome/checkpointing/app/CheckpointingComposition.scala#L372-L383): instead of the locking, create transactions.

enriquerodbe (approved):

aakoshh:
@dmitry-worker do you want to check out why your commits aren't signed and push new versions? Or should I merge this for you?

